PR #18, opened 2021-01-27
feat: Default prefix for ESLMedia event prefixes

ala-n:
The default prefix for ESLMedia event prefixes updated to standardize approach in ESL

BREAKING CHANGE:
 ESLMedia events use 'media:' prefix instead of 'esl:media:'


dshovchko (approved):

ala-n:
Declined due event name strategy change

PR #20, opened 2021-01-29
feat: introduce new BasePopupGroupManager based on events

julia-murashko:


src/modules/esl-base-popup/core/esl-base-popup.ts
```diff
@@ -43,38 +40,36 @@ export class ESLBasePopup extends ESLBaseElement {
   @boolAttr() public closeOnEsc: boolean;
   @boolAttr() public closeOnOutsideAction: boolean;
 
-  @jsonAttr<PopupActionParams>({defaultValue: {silent: true, force: true, initiator: 'init'}})
```
ala-n, line 46: We can remove `silent` from the default object literal in case it is false.

src/modules/esl-base-popup/core/esl-base-popup.ts
```diff
@@ -43,38 +40,36 @@ export class ESLBasePopup extends ESLBaseElement {
...
   protected attributeChangedCallback(attrName: string, oldVal: string, newVal: string) {
     if (!this.connected || newVal === oldVal) return;
     switch (attrName) {
       case 'open':
         this.toggle(this.open, Object.assign({initiator: 'attribute'}, this.defaultParams));
         break;
       case 'group':
-        oldVal && ESLBasePopupGroup.unregister(this, oldVal);
-        newVal && ESLBasePopupGroup.register(this, newVal);
+        this.$$fire('change:group',  {
```
ala-n, line 55: Event is 
```
 bubbles: true,
          composed: true,
          cancelable: true,
```
by default, and it merges all settings

ala-n (approved):

src/modules/esl-base-popup/core/esl-base-popup.ts
julia-murashko: Fixed. Thanks!

src/modules/esl-base-popup/core/esl-base-popup.ts
julia-murashko: Fixed. Thanks!

dshovchko (approved):

src/modules/esl-panel/core/esl-panel.ts
```diff
@@ -24,7 +24,7 @@ export class ESLPanel extends ESLBasePopup {
   @boolAttr() public isAccordion: boolean;
   @boolAttr() public startAnimation: boolean;
 
-  @jsonAttr<PanelActionParams>({defaultValue: {silent: true, force: true, initiator: 'init', noAnimation: true}})
+  @jsonAttr<PanelActionParams>({defaultValue: {silent: false, force: true, initiator: 'init', noAnimation: true}})
```
ala-n, line 27: ???

package-lock.json
```diff
@@ -11639,7 +11639,7 @@
       "dependencies": {
         "JSONStream": {
           "version": "1.3.5",
-          "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.5.tgz",
+          "resolved": false,
```
ala-n, line 11642: Do we really need this update? What was the root cause?

src/modules/esl-panel/core/esl-panel-stack.ts
```diff
@@ -52,6 +52,7 @@ export class ESLPanelStack extends ESLBaseElement {
 
   @bind
   protected _onShowPanel(e: CustomEvent) {
+    if (!e.detail.open) return; // TODO check
```
ala-n, line 55: That one causing error on Tabs component

ala-n (changes requested):

src/modules/esl-base-popup/core/esl-base-popup-group-manager.ts
```diff
@@ -0,0 +1,129 @@
+import {ESLBasePopup} from './esl-base-popup';
+import {bind} from '../../esl-utils/decorators/bind';
+import {ExportNs} from '../../esl-utils/environment/export-ns';
+import {memoize} from '../../esl-utils/decorators/memoize';
+
+@ExportNs('BasePopupGroupManager')
+export class ESLBasePopupGroupManager {
```
ala-n, line 7: Separate request: shadow DOM break group functionality

src/modules/esl-panel/core/esl-panel-stack.ts
ala-n: @julia-murashko, that was reverted as it is a blocker to pass regression.

ala-n (approved):

PR #87, opened 2021-03-15
feat: add ability to cancel animation

julia-murashko:


ala-n (changes requested):
Discussed in PM
Use case and solution approach transformed.

ala-n (approved):

PR #137, opened 2021-04-12
feat: extend Scrollable Tabs to support center position

julia-murashko:


dshovchko (approved):

src/modules/esl-tab/core/esl-tabs.ts
```diff
@@ -19,7 +19,7 @@ export class ESLTabs extends ESLBaseElement {
   public static is = 'esl-tabs';
 
   /** Marker to enable scrollable mode */
```
ala-n, line 21: Docs update with possible values

yadamskaya (approved):

src/modules/esl-tab/core/esl-tabs.ts
ala-n: Also, it sounds like we need the default value
because we should support 3 states (disabled, centered, side alignment)
and we should not break the existing API:
scrollable = false (no attribute) -> disabled
scrollable = true (attribute with no value) -> side aligned scrollable

src/modules/esl-tab/core/esl-tabs.ts
```diff
@@ -93,26 +93,30 @@ export class ESLTabs extends ESLBaseElement {
     const areaRect = $scrollableTarget.getBoundingClientRect();
     const itemRect = $trigger.getBoundingClientRect();
 
-    let shift = 0;
+    $scrollableTarget.scrollBy({
+      left: this.calcScrollOffset(itemRect, areaRect),
```
ala-n, line 97: 👍🏻

ala-n (approved):

src/modules/esl-tab/core/esl-tabs.ts
julia-murashko: Thanks!

PR #150, opened 2021-04-16
ESLTabs: adjust scrollable attribute

julia-murashko:


src/modules/esl-tab/README.md
```diff
@@ -11,9 +11,9 @@ ESLToggleable or, usually, ESLPanel.
 
 - `container` - inner element to contain {@link ESLTab} collection. Will be scrolled in a scrollable
 
-- `scrollable` - scrollable mode. The following values are supported:
-    - 'none' or not defined -  scroll behavior is disabled;
-    - 'centered' - scroll behavior is enabled, tab is center-aligned;
+- `scrollable` - scrollable mode. Supported types for different breakpoints ('disabled' by default) :
+    - 'disabled' or not defined -  scroll behavior is disabled;
+    - 'center' - scroll behavior is enabled, tab is center-aligned;
     - empty or unsupported value - scroll behavior is enabled, tab is side-aligned;
```
ala-n, line 17: 'side' with a note that it is the default for empty or unsupported value

src/modules/esl-tab/core/esl-tabs.ts
```diff
@@ -72,8 +87,8 @@ export class ESLTabs extends ESLBaseElement {
   }
 
   /** Active {@link ESLTab} item */
-  public get $current(): ESLTab | null {
-    return this.$tabs.find((el) => el.active) || null;
+  public get $current(): ESLTab | undefined {
```
ala-n, line 90: Do we really need that change?
`undefined` is the value that is similar to the unsupported key, while the null value indicates that it is supported key but there is no current tub right now

src/modules/esl-tab/core/esl-tabs.ts
```diff
@@ -181,4 +196,44 @@ export class ESLTabs extends ESLBaseElement {
...
+  protected updateScrollableType() {
+    ESLTabs.supportedScrollableTypes.forEach((type) => {
+      CSSUtil.toggleClsTo(this, `${type}-alignment`, this.currentScrollableType === type);
+    });
+    this.fitToViewport(this.$current);
+
+    if (this.currentScrollableType === 'disabled') {
+      this.unbindScrollableEvents();
+    } else {
+      this.bindScrollableEvents();
```
ala-n, line 236: Please remove the same binding from the connected callback, its better to have one flow to initiate this,

src/modules/esl-tab/core/esl-tabs.ts
```diff
@@ -18,24 +20,35 @@ import {debounce} from '../../esl-utils/async/debounce';
...
+      this.updateScrollableType();
+    }
+  }
+
   protected connectedCallback() {
     super.connectedCallback();
 
-    if (this.isScrollable) {
-      this.bindScrollableEvents();
-      this.updateScroll();
```
ala-n, line 37: It looks like the `this._deferredFitToViewport(this.$current, 'auto');` part from the update scroll is missed

ala-n (approved):

PR #161, opened 2021-04-23
ESL Tabs: adjust dynamic change of scrollable attr

julia-murashko:


src/modules/esl-tab/core/esl-tabs.scrollable.less
```diff
@@ -15,4 +15,8 @@ esl-tabs[scrollable] {
   .arrow-prev, .arrow-next {
     flex: 0 0 auto;
   }
+
+  .esl-scrollable-content {
```
ala-n: `.esl-scrollable-content` is the common modifier class so we can not affect all of the elements with a such marker
Consider direct child check '> .esl-scrollable-content' and/or .esl-tabs-container clarification or manual control over the hidden scroll

julia-murashko: Thanks!

ala-n (approved):

PR #193, opened 2021-05-17
esl-panel-group: support mode="open" to make all panels active

julia-murashko:
Beta version of implementation


ala-n (approved):

ala-n:
Approved for beta implementation.
May and will be changed before stable release

dshovchko (approved):

PR #216, opened 2021-05-28
feat:  add omit function

nattallius:


ala-n:
Discussed with a core team, the result proposed strategy is the following:
`copyDefinedKeys` is not going to be deprecated but should be defined as an alias  (PAF) for `omit` call with `(key, val) => val !== void 0` predicate.
'omit' is approved to be released in v2 stable release.

src/modules/esl-utils/misc/object.ts
```diff
@@ -40,6 +43,16 @@ export function copyDefinedKeys<T>(obj?: T): Partial<T> {
   return result;
 }
 
+/** Makes a copy of obj with properties satisfying the predicate */
+export function omit(obj: any, predicate: (key: string, value?: any) => boolean): Record<string, any> {
```
ala-n, line 47: Try to use template type instead see `copyDefinedKeys` definition above

ala-n, line 47: There is also no need to define value: any as optional because it will be available anyway

YanaBr (approved):

andreybelous (approved):

ala-n (approved):

src/modules/esl-utils/misc/object.ts
```diff
@@ -33,13 +34,24 @@ export function defined<T>(...params: T[]) {
 
 /** Makes a flat copy without undefined keys */
 export function copyDefinedKeys<T>(obj?: T): Partial<T> {
+  return copy(obj || {}, (key, value) => value !== void 0);
+}
+
+/** Makes a plain copy of obj with properties satisfying the predicate
+ * If no predicate provided copies all own properties */
+export function copy<T>(obj: T, predicate: CopyPredicate  = () => true): Partial<T> {
```
ala-n, line 42: Super-cosmetic: but it's better to place generic implementation fist, and the consumers (omit, copyDefKeys) after

src/modules/esl-utils/misc/object.ts
```diff
@@ -1,5 +1,6 @@
 export const isObject = (obj: any): obj is Record<string, any> => obj && typeof obj === 'object';
 export const isObjectLike = (obj: any) => isObject(obj) || typeof obj === 'function';
+export type CopyPredicate = (key: string, value: any) => boolean;
```
ala-n, line 3: Super-cosmetic: space to separate type exports from real objects

ala-n (commented):
Also, update esl-utils README.md, please

PS: merge PR with squash to clearly describe the resulting change for changelog

src/modules/esl-utils/misc/object.ts
```diff
@@ -31,15 +33,26 @@ export function defined<T>(...params: T[]) {
...
 }
 
+/** Makes a flat copy without undefined keys */
+export function copyDefinedKeys<T>(obj?: T): Partial<T> {
+  return copy(obj || {}, (key, value) => value !== void 0);
+}
+
+/** Omit copying provided properties from object */
+export function omit<T>(obj: T, keys: string[]) {
+  return copy(obj, key=> !keys.includes(key));
```
dshovchko, line 53: typo: missing space

dshovchko (approved):

src/modules/esl-utils/misc/object.ts
nattallius: Thank you!

PR #218, opened 2021-05-31
fix(esl-select):  fix height empty space appeared with little amount of list items

nattallius:


abarmina (approved):

pages/views/pages/forms/esl-select.html
```diff
@@ -582,59 +582,6 @@
               <option value="#72A0C1">Air superiority blue</option>
               <option value="#EDEAE0">Alabaster</option>
               <option value="#EFDECD">Almond</option>
-              <option value="#E52B50">Amaranth</option>
```
ala-n, line 585: Please better add an additional example instead of updating an existing one

src/modules/esl-forms/esl-select-list/core/esl-select-list.less
```diff
@@ -8,8 +8,8 @@
   border: 1px solid @esl-select-list-secondary;
 
   min-height: 50px;
-  height: 400px;
-  max-height: 100%;
+  height: auto;
+  max-height: 50vh;
```
ala-n, line 12: This limitation basically sounds more like a dropdown one.
So possibly we can have some fixed limit here, and override for a list inside select-dropdown

ala-n (approved):

src/modules/esl-forms/esl-select-list/core/esl-select-list.less
nattallius: Yes, but if the list itself has few items ( I mean with no dropdown), there will be empty space just in the list. Is it ok, or I am missing sth?

ala-n: I think that is ok for a simple list.
But I'm just asking about fixed limitations like 400px for the max-height in case of the mostly "in-flow" element - list, while its usage in the dropdown will depend on the viewport height.

ala-n (approved):

julia-murashko (approved):

PR #244, opened 2021-06-11
fix: pseudo scroll locker styles fix

ala-n:


Sisha3342 (approved):

yadamskaya (approved):

src/modules/esl-utils/dom/scroll.less
```diff
@@ -40,10 +40,12 @@ html.esl-disable-scroll {
       right: 0;
       bottom: 0;
       overflow-y: scroll;
+      // Pseudo scroll should be visible above the content
+      z-index: 100000;
```
dshovchko, line 44: I've seen some have great values. For example, AddThis has from 100000 and above.

dshovchko (approved):

src/modules/esl-utils/dom/scroll.less
ala-n: Let's go with a little compromise, as you suggested: 999999 :) Anyway, that's just a pseudo scroll and just a CSS that is overridable by consumers.
Thanks!

julia-murashko (approved):

PR #249, opened 2021-06-14
fix: add polyfill prefix '/esl/' for correct url

abarmina:


nattallius (approved):

pages/views-11ty/layouts/basic.njk
```diff
@@ -11,7 +11,7 @@
       } else if (!('getRootNode' in Element.prototype) || !window.customElements || !window.IntersectionObserver) {
         version = 'light';
       }
-      (version) && document.write('<script type="text/javascript" src="/bundles/polyfill-' + version + '.js"><\/script>');
+      (version) && document.write(`<script type="text/javascript" src="{{ '/bundles/polyfill-${version}.js' | url }}"><\/script>`);
```
ala-n, line 14: Template strings are part of ES6 syntax, so there is no reason to load polyfill for IE with the way that IE doesn't recognize ))

ala-n (changes requested):

pages/views-11ty/layouts/basic.njk
abarmina: it's funny! thanks, I didn't think about it

ala-n (approved):

PR #250, opened 2021-06-14
fix: additional components build

abarmina:


Sisha3342 (approved):

package.json
```diff
@@ -36,7 +36,10 @@
     "modules/draft/esl-carousel/core/view/*.js",
     "src/modules/draft/esl-carousel/core/view/*.ts",
     "polyfills/**/*.js",
-    "src/polyfills/**/*.ts"
+    "src/polyfills/**/*.ts",
+    "pages/src/common/*.ts",
```
ala-n, line 40: You can basically move everything under the pages/src/**/*.ts under side effects to not overload library sideEffects with auxiliary paths.

ala-n (approved):

dshovchko (approved):

YanaBr (approved):

nattallius (approved):

package.json
abarmina: good point

ala-n (approved):

PR #352, opened 2021-07-29
feat: ESLPopup positioning

dshovchko:
That pull request contains the following changes:
- added calculations for new positions of popup: 'left', 'right' and 'bottom';
- added scrolling tracking mechanism for cases when popup trigger is inside scroll parent. The number of such scroll parents is unlimited, all are tracking;
- added intersection tracking mechanism for cases when popup trigger is going to out of the visible area;
- added a mechanism for detecting the direction of scrolling, as well as on which of the major or minor axes scrolling is performed, and the associated mechanism for mirroring the position of the popup and hiding it;
- optimized popup position update. Now it takes on average 0.2msec for the update after scroll and 2-3msec during the first showing (the same productivity on both Intel Core i7-8850H CPU and AMD A8-7680 CPU);
- added new dom helpers to esl-utils: element, rect, and window;
- updated dom scroll helper with methods for getting and checking scroll parents.


src/modules/esl-popup/core/esl-popup-position.ts
```diff
@@ -0,0 +1,196 @@
...
+ * @param cfg - popup position config
+ * @param rect - popup position rect
+ * @param arrow - arrow position value
+ * */
+function fitByMajorAxis(cfg: PopupPositionConfig, rect: Rect, arrow: ArrowPositionValue) {
+  if (cfg.behavior !== 'fit') return;
+
+  let isMirrored = false;
+  switch (cfg.position) {
+    case 'bottom':
```
ala-n, line 83: DIscussable super cosmetic:
It may look compacter in case we replace the switch with extra conditions. 
But not sure here, so it's definitely rejectable super cosmetic.

src/modules/esl-popup/core/esl-popup.ts
```diff
@@ -41,24 +56,31 @@ export class ESLPopup extends ESLToggleable {
   @prop() public closeOnOutsideAction = true;
 
   public connectedCallback() {
-    this.$arrow = this.querySelector('.esl-popup-arrow');
-
     super.connectedCallback();
+
+    setTimeout(() => {
+      this.$arrow = this.querySelector('span.esl-popup-arrow');
+    });
```
ala-n, line 63: Why was it moved to a separate task?
Mb we need to mark the whole connected callback with @ready decorator instead?

src/modules/esl-popup/core/esl-popup.ts
```diff
@@ -74,88 +96,121 @@ export class ESLPopup extends ESLToggleable {
...
+      threshold: [...Array(9).keys()].map((x) => x/8)
+    } as IntersectionObserverInit;
 
-    const right = this._windowWidth - (left + this.offsetWidth);
-    if (this.behavior === 'fit' && right < this._offsetWindow) {
-      arrowAdjust -= right - this._offsetWindow;
-      left += right - this._offsetWindow;
-    }
-    const arrowLeft = this.clientWidth / 2 + arrowAdjust;
+    const observer = new IntersectionObserver(this.onActivatorIntersection, options);
```
ala-n, line 162: It will make sense to move IObserver to a statically defined (not compulsory the class static member, but at least make it single across popup instances)
At least, in case we are not going to declare options dependently of instance
IObserver prefers to be shared across consumers instead of being created separately for each instance. It will simplify unsubscription by the way...

src/modules/esl-popup/core/esl-popup.ts
```diff
@@ -74,88 +96,121 @@ export class ESLPopup extends ESLToggleable {
...
+      };
+    });
 
-    if (this.behavior === 'fit' && left < this._offsetWindow) {
-      arrowAdjust += left - this._offsetWindow;
-      left = this._offsetWindow;
-    }
+    const options = {
+      rootMargin: '0px',
+      threshold: [...Array(9).keys()].map((x) => x/8)
```
ala-n, line 159: Do we really use all of the breakpoints here? Big amount of breakpoints can affect performance.

src/modules/esl-utils/dom/element.ts
```diff
@@ -0,0 +1,41 @@
+import {getWindow} from './window';
+
+/**
+ * Get an object containing the values of all CSS properties of an element.
+ * @param element - element for which to get the computed style
+ * */
+export function getComputedStyle(element: Element): CSSStyleDeclaration {
+  return getWindow(element).getComputedStyle(element);
```
ala-n, line 8: Do we know why do we need to make sure of the window object here?
It looks like an electron environment support shim basically or some super strange situation with iframes from the same domain.

src/modules/esl-utils/dom/rect.ts
```diff
@@ -0,0 +1,113 @@
...
+   * The Y coordinate of the Rect's origin (top-left corner of the rectangle).
+   * */
+  public y = 0;
+  /**
+   * The width of the Rect.
+   * */
+  public width = 0;
+  /**
+   * The height of the Rect.
+   * */
```
ala-n, line 19: Cosmetic: it's preferable to use compact one-line comments. Basically no problems with the just an agreement in bounds of the project
`/** One line comment */`

src/modules/esl-utils/dom/rect.ts
```diff
@@ -0,0 +1,113 @@
...
+    return new this(
+      rect.x ?? 0,
+      rect.y ?? 0,
+      rect.width ?? 0,
+      rect.height ?? 0,
+    );
+  }
+
+  public constructor(x?: number, y?: number, width?: number, height?: number) {
+    if (x != null) {
```
ala-n, line 36: It looks like if we make it undefined or null it will be 0.
So propose to make 
```
this.x = x || 0;
this.y = y || 0;
this.width= width || 0;
this.height = height || 0;
```
in the constructor and then get rid of properties defaults (as they will add additional transpiled code)
and get rid of all checks in method `from` above 
`return new this(rect.x, rect.y, rect.width, rect.height); `

ala-n (approved):

src/modules/esl-popup/core/esl-popup.ts
dshovchko: It's a "horrible piece of shit" (c).
I added this crutch and completely forgot.
Fixed

src/modules/esl-popup/core/esl-popup.ts
dshovchko: Quite a complicated question. Either we improve performance or we improve the smoothness and animations.
I had planned to take the issue of the behavior of the popup at the crossing points and leaving the window. But will make it a separate PR.
I guess need to experiment with rootMargin and thresholds to get a compromise result to satisfy performance and smoothness.

src/modules/esl-popup/core/esl-popup.ts
dshovchko: Also, not everything is clear here.
For example, as a future feature... Is it possible for popups to set different rootMargin? It was just that this item was raised during the discussions.
Let's not rush for now and will discuss it?

ala-n: Yep, of cose. Thats just a point to discuss in future.

julia-murashko (approved):

src/modules/esl-utils/dom/rect.ts
dshovchko: Fixed

src/modules/esl-utils/dom/rect.ts
dshovchko: done

src/modules/esl-utils/dom/element.ts
dshovchko: Got rid of this method.

src/modules/esl-popup/core/esl-popup-position.ts
dshovchko: This is unlikely to give an improvement in this particular place.
But I did manage to do this trick later on this code.

andreybelous (approved):

NastaLeo (approved):

pages/views/_layouts/html5.njk
```diff
@@ -8,6 +8,7 @@
 </head>
 <body>
 {% block body %}{% endblock %}
+{% block extra_body %}{% endblock %}
```
abarmina, line 11: maybe add this block to basic.njk, line 42? it will be more transparent logic of extensions and the code will be before adding scripts

abarmina (approved):

pages/views/_layouts/html5.njk
dshovchko: OK, thanks!

Sisha3342 (approved):

yadamskaya (approved):

YanaBr (approved):

PR #356, opened 2021-08-02
feature(markdown): rendering certain parts of markdown to page

abarmina:
- ability to leave a certain part of markdown from page rendering (using anchor links)
- source code from markdown is displayed correctly on a page
- added syntax highlighting for source code
- minor style edits


package.json
```diff
@@ -68,7 +68,8 @@
   "dependencies": {
     "@types/resize-observer-browser": "0.1.6",
     "@types/video.js": "7.3.24",
-    "@types/youtube": "0.0.44"
+    "@types/youtube": "0.0.44",
+    "highlight.js": "^11.1.0"
```
ala-n, line 72: Not a library dependency. Move to devDependencies, please.

abarmina: Oh yeah, thanks!

.eleventy.js
```diff
@@ -1,5 +1,17 @@
...
+  html: true,
+  highlight: function (str, lang) {
+    if (lang && hljs.getLanguage(lang)) {
+      try {
+        return hljs.highlight(str, { language: lang }).value;
+      } catch (__) { }
+    }
+    return '';
+  }
+});
```
dshovchko, line 14: Maybe markdown-it-highlightjs use instead?

abarmina: I had difficulties connecting it, today I will try again and come back with an answer

ala-n: I don't think this is a good idea. It's a super simple library, but it does not have enough support as I see.
So prefer to use our own code that we have access to instead of using third libraries.

abarmina: > Maybe markdown-it-highlightjs use instead?

Eleventy has its own syntax highlighting, but since we use shortcode {% markdown %}, we cannot use it. I  think highlight.js is the best solution now

ala-n (approved):

pages/views/_data/md-render.js
```diff
@@ -0,0 +1,56 @@
...
+      const content = await parseFile(filePath);
+
+      if (!startSel) return MDRenderer.wrapContent(content);
+
+      const { window } = new JSDOM(content);
+      const startAnchor = window.document.querySelector(`a[name='${startSel}']`);
+      const endAnchor = window.document.querySelector(`a[name='${endSel}']`);
+
+      let node = unwrapAnchor(startAnchor);
+      console.log(node, '')
```
dshovchko, line 41: Is it debugging left?

dshovchko (approved):

pages/views/_data/md-render.js
abarmina: Oh, yep, thanks

Issue #292, opened 2021-06-25
GH Pages: setup a sitemap generator


ala-n:
Setup sitemap generation template for ESL demo site pages

PR #362, opened 2021-08-03
GH Pages: setup a sitemap generator #292

NastaLeo:
Created **sitemap.njk** (includes info about all urls, date of its last modification and its priority of (in our case, homepage - 1,0, others - 0,9), **robots.njk** (contains link to our sitemap file).


.eleventy.js
```diff
@@ -15,6 +15,7 @@ module.exports = (config) => {
     return [...values].sort((a, b) => a.data.name.localeCompare(b.data.name));
   });
 
+
```
dshovchko, line 18: cosmetic: extra line

pages/views/robots.njk
```diff
@@ -0,0 +1,8 @@
+---
+permalink: "/robots.txt"
+eleventyExcludeFromCollections: true
+---
+User-agent: *
+Allow: /
+
+Sitemap: {{ site.url }}/sitemap.xml
```
dshovchko, line 8: It seems to me that robots.txt must contain a new line at the end of the file. (unlike the sitemap.xml file)

pages/views/sitemap.njk
```diff
@@ -0,0 +1,17 @@
...
+---
+
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+{% for page in collections.all %}
+{% if page.data.ignore != true %}
+  <url>
+    <loc>{{ site.url }}{{ page.url.replace(r/\/$/, ".html") if page.data.tags else page.url }}</loc>
+    <lastmod>{{ page.date.toISOString() }}</lastmod>
+    <priority>{{ "1.00" if page.url == "/" else "0.90" }}</priority>
```
dshovchko, line 13: It is enough to use values up to one decimal place.
Moreover, it is enough to specify the priority for only one root page. From the resulting file, it will be clear to the search engine crawler that you give priority only to the root page. The rest of the pages will be considered equivalent, so you can use the default value for them and therefore not set a value at all. In fact, this priority will not affect the search results in any way. Only for the order in which the pages are scanned.

dshovchko (approved):

pages/views/sitemap.njk
```diff
@@ -0,0 +1,17 @@
+---
+permalink: "/sitemap.xml"
+eleventyExcludeFromCollections: true
+---
+
```
ala-n, line 5: Cosmetic: please remove empty space here, it is better to have
`<?xml version="1.0" encoding="UTF-8"?>`
at the very beginning of the xml document (some riders can assume that as a warning)

pages/views/sitemap.njk
```diff
@@ -0,0 +1,17 @@
...
+{% for page in collections.all %}
+{% if page.data.ignore != true %}
+  <url>
+    <loc>{{ site.url }}{{ page.url.replace(r/\/$/, ".html") if page.data.tags else page.url }}</loc>
+    <lastmod>{{ page.date.toISOString() }}</lastmod>
+    <priority>{{ "1.00" if page.url == "/" else "0.90" }}</priority>
+  </url>
+{% endif %}
+{% endfor %}
+<urlset>
```
ala-n, line 17: Major: Incorrect closing tag (opening tag instead of closing one) + empty end-line

pages/views/sitemap.njk
```diff
@@ -0,0 +1,17 @@
+---
+permalink: "/sitemap.xml"
+eleventyExcludeFromCollections: true
+---
+
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+{% for page in collections.all %}
```
ala-n, line 8: Major: drafts collection should be excluded from the sitemap

pages/views/sitemap.njk
NastaLeo: Done

pages/views/sitemap.njk
```diff
@@ -0,0 +1,19 @@
+---
+permalink: "/sitemap.xml"
+eleventyExcludeFromCollections: true
+---
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+  {%- for page in collections.all %}
+  {%- if page.data.ignore != true %}
```
abarmina, line 8: if we only want to exclude drafts collection, mb we can use this line `{%- if not (page.data.tags == 'drafts') %}` (and we don't need to add to drafts `ignore: true`)

pages/views/404.njk
```diff
@@ -3,6 +3,7 @@ layout: basic
 title: 😯 404 😯
 containerCls: container
 permalink: 404.html
+ignore: true
```
abarmina, line 6: I would suggest using here also `eleventyExcludeFromCollections: true`

pages/views/sitemap.njk
NastaLeo: It's really perfect syntax. Thanks

abarmina (approved):

ala-n (approved):

julia-murashko (approved):

Issue #332, opened 2021-07-15
Bug: TS types are missing for document.createElement default ESL tag names


ala-n:
As an ESL consumer, I want to have an optional type extension for document.createElement with default ESL tag names.

**Example**
```ts
const el: ESLToggleable = document.createElement('esl-toggleable'); // type error
```

**Expected behavior**
```ts
const el: ESLToggleable = document.createElement('esl-toggleable');
```

**Environment**
- TypeScript

PR #366, opened 2021-08-05
Bug: TS types are missing for document.createElement default ESL tag names #332

NastaLeo:
global declaration of elements types


package.json
```diff
@@ -30,7 +30,7 @@
   "types": "modules/all.d.ts",
   "sideEffects": [
     "modules/lib.js",
```
ala-n, line 32: modules/namespace.js

ala-n (approved):

Sisha3342 (approved):

ala-n:
Sorry, will use this pr to resolve typings issues.
The typing doesn't seem to work correctly with the strategy we choose, so update with distributed typings.
Also, there were some lint fixes, and namespace types rework.

Team, please be aware of the new type of placement.

PR #393, opened 2021-08-17
Feat/add new format

Anna-MariiaPetryk:


ala-n:
Reqires squash to merge changes with the correct commit description

src/modules/esl-utils/misc/format.ts
```diff
@@ -1,4 +1,4 @@
-import {get} from './object';
+import { get } from './object';
```
ala-n, line 1: Spaces should be warned by eslint
Something went wrong with our current eslint checker ...

src/modules/esl-utils/misc/format.ts
```diff
@@ -49,10 +49,18 @@ export function evaluate(str: string, defaultValue?: any): any {
...
     const val = get(source, key);
     return val === undefined ? match : val;
   });
 }
+
+
+
+
+
+
```
ala-n, line 66: Extra spaces are also should not be allowed

ala-n (commented):
I have the following news :)
(1) Changes looks good for me, (2) but they are not eslint rules compliant (3) should be... but looks like we break eslint (

PR #395, opened 2021-08-17
Feat/add arr css class utils

Anna-MariiaPetryk:


ala-n (changes requested):
Commit message needs to be reword

ala-n (approved):
Use squash to merge changes with the correct commit discription

PR #450, opened 2021-09-02
fix: incorrect moment of change outside action listener state

ala-n:


yadamskaya (approved):

docs/contribute/commit.md
```diff
@@ -49,6 +49,10 @@ Type is a required part of the message, and it is limited by the following value
 
 Type should be in a lowercase.
 
+### Force minor version update
+
+You can increase importance of the patch changes to the minor using the `MAJOR` marker in the message.
```
dshovchko, line 54: This confused me. I don't understand why it is necessary to specify a `MAJOR` marker for minor changes.

dshovchko (approved):

julia-murashko (approved):

docs/contribute/commit.md
ala-n: Semiver is [Major].[Minor].[Patch] 

In case make a fix o refactoring - PATCH part increasing
In case we do a feature - MINOR part is increasing
In case we mark BREAKING CHANGES - then independently of the type of fix the MAJOR version increasing
 
 Here are some additional mechanisms to make a fix or refactoring Minor changes instead of Patch.
 I know that `MAJOR` is a little bit confusing, but here it is meant to be native meaning. Sure we can replace this keyword with something more neutral...

PR #455, opened 2021-09-06
feat: add intersection, union,  complement, fullIntersection, array utility methods

Anna-MariiaPetryk:
- add new utilities to the set group due to the uniqueness requirement/result for added utilities


src/modules/esl-utils/misc/array.ts
```diff
@@ -29,13 +29,34 @@ export const uniq = <T> (arr: T[]): T[] => {
...
   return arr;
 }
+
+/** Create an array of unique values from two arrays*/
+export function intersection(a: any, b: any): any[] {
+  const arr = new Set([...a, ...b]);
+  return [...arr];
+}
+
+/** Create an array from two arrays*/
```
ala-n, line 49: Not enough for documentation the current one says nothing but the method signature
`/** Create an array with the unique values from each of the passed arays */`

src/modules/esl-utils/misc/array.ts
```diff
@@ -29,13 +29,34 @@ export const uniq = <T> (arr: T[]): T[] => {
...
+export function union(a: any, b: any): any[] {
+  return [...a, ...b];
```
ala-n, line 51: 1) Incorrect implementation sounds like the current intersection implementation is union:
`expect(union([1, 2], [1, 2, 3])).toEqual([1, 2, 3])`
2) It will be better to make method support more then array pair, and accept unlimited number of initial arrays

src/modules/esl-utils/misc/array.ts
```diff
@@ -29,13 +29,34 @@ export const uniq = <T> (arr: T[]): T[] => {
...
+}
+
+/** Creates an array of unique values from the first array that are not present in the second array*/
```
ala-n, line 54: Cosmetic: 
1) ending space expected:  `array */`
2) the current implementation doesn't require uniqueness so it is better to remove that declaration in the doc
3) `expect(complement([1,2,3,1,2,3], [1,2,3])).toEqual([1,2,3])`

ala-n (changes requested):
Need to request updates:
- please see mentioned requirements in the comment
- update test set acordingly

src/modules/esl-utils/misc/array.ts
```diff
@@ -29,13 +29,44 @@ export const uniq = <T> (arr: T[]): T[] => {
...
+  if (rest.length > 1) return complement(a, complement(b, rest));
+  const setB = new Set(b);
+  return a.filter(element => !setB.has(element));
+}
+
+
+/** Check for elements from array B in array A */
+export function fullIntersection(a: any[], b: any[]): boolean {
+  const setA = new Set(a);
+  return a.length === 0 && b.length === 0 ? true : !b.filter((item: any) => !setA.has(item)).length;
```
ala-n, line 71: 1) Move simple condition upper before the method creates a set.
2) Use `some` or `every` instead of checking the filter result (filter creates a new array and check each item of the array, while the predicates `some`/`every`  check items while it's reasonable from boolean algebra)

src/modules/esl-utils/misc/array.ts
```diff
@@ -29,13 +29,44 @@ export const uniq = <T> (arr: T[]): T[] => {
...
+/** Creates an array of unique values from the first array that are not present in the other arrays */
+export function complement<T>(...rest: T[][]): T[];
+export function complement(a: any[], b: any[], ...rest: any[][]): any[] {
+  if (rest.length > 1) return complement(a, complement(b, rest));
+  const setB = new Set(b);
+  return a.filter(element => !setB.has(element));
+}
+
+
+/** Check for elements from array B in array A */
```
ala-n, line 68: `Check if the passed arrays have a full intersection (A ⋂ B = A || A ⋂ B = B)`

src/modules/esl-utils/misc/array.ts
```diff
@@ -29,13 +29,44 @@ export const uniq = <T> (arr: T[]): T[] => {
...
+  const set = new Set();
+  rest.forEach(item => item.forEach(i => set.add(i)));
+  const result: any[] = [];
+  set.forEach(value => result.push(value));
+  return result;
+}
+
+/** Creates an array of unique values from the first array that are not present in the other arrays */
+export function complement<T>(...rest: T[][]): T[];
+export function complement(a: any[], b: any[], ...rest: any[][]): any[] {
```
ala-n, line 61: Cosmetic: it will be better to make original sydnature strict (with generic type) until it's really necessary to disable type check with `any`

ala-n: She same for othe r methods

src/modules/esl-utils/misc/array.ts
```diff
@@ -29,13 +29,44 @@ export const uniq = <T> (arr: T[]): T[] => {
...
+
+/** Create an array of unique values that presented in each of the passed arrays */
+export function intersection<T>(...rest: T[][]): T[];
+export function intersection(a: any[], b: any[], ...rest: any[][]): any[] {
+  if (rest.length) return intersection(a, intersection(b, ...rest));
+  return a.filter(Set.prototype.has, new Set(b));
+}
+
+/** Create an array with the unique values from each of the passed arays */
+export function union(...rest: any[][]): any[] {
```
ala-n, line 51: Strict types ?

src/modules/esl-utils/misc/test/array.test.ts
```diff
@@ -45,3 +45,40 @@ describe('misc/array helper tests', () => {
     expect(range(9, (x) => x / 8)).toEqual([...Array(9).keys()].map((x) => x / 8))
   });
 });
+
+test('intersection', () => {
```
ala-n, line 49: Please also check edge cases:
Currently method signature accepts 0 or 1 argument, find the way to disallow such cases or make intersect return an empty array or the first argument.
The same applicable to other methods

ala-n (changes requested):

ala-n (approved):

abarmina (approved):

julia-murashko (approved):

yadamskaya (approved):

PR #490, opened 2021-09-21
feat: add ability to prevent scroll on focus

Anna-MariiaPetryk:


dshovchko (approved):

src/modules/esl-a11y-group/README.md
```diff
@@ -1,20 +1,22 @@
...
 ESLA11yGroup - a helper custom element that adds a11y group behavior to targets.
-ESLA11yGroup tracks keyboard to loop focus of the targeting elements (with arrow keyboard keys navigation). 
+ESLA11yGroup tracks keyboard to loop focus of the targeting elements (with arrow keyboard keys navigation).
 
 ### Attributes / Properties
 
-- `targets` - property to declare targets for the group using [TraversingQuery](https://exadel-inc.github.io/esl/utils/esl-traversing-query/) syntax. Use group parent element as a base element.
+- `targets` - property to declare targets for the group using [TraversingQuery](https://exadel-inc.github.io/esl/utils/esl-traversing-query/) syntax. Use group parent element as a base element
 - `activate-selected` - activate the element using `click` event on selection change via esl-a11y-group
+- `prevent-scroll` - prevants scroll when target receive focus
```
abarmina, line 16: mb typo `prevents` and `receives`
if we decided to use "s" endings of the verbs, maybe change activate-selected `activate` and `Use` in targets? @ala-n, what do you think?

abarmina (approved):

src/modules/esl-a11y-group/README.md
ala-n: Yep, I basically think that's a good point, we can try to fix docs according to our current agreements in case we touching something nearby or related to our scope changes.

ala-n (approved):

PR #572, opened 2021-10-18
chore: add new svg images

abarmina:
# PR Details

Add images in SVG format for redesign


abarmina:
> We need to have a `viewBox` attribute defined instead of `width/height` for SVGs in order to correctly handle inline assets size using CSS. PS: pay attention to GitHiub common icon that most likely have an incorrect size right now on that branch.

oh, there was a `viewBox` attribute, but it was removed by the optimizer, apparently. I'll fix it.
about the GitHiub icon - I'll look at it

PR #610, opened 2021-11-02
Benefits component

Anna-MariiaPetryk:
# PR Details
Benefits component implementation for the landing page.


pages/views/_data/site.json
```diff
@@ -4,5 +4,37 @@
   "github": {
     "url": "https://github.com/exadel-inc/esl",
     "srcUrl": "https://github.com/exadel-inc/esl/tree/main/"
-  }
+  },
+  "benefit" : [
```
abarmina, line 8: I suggest moving this to a separate file in the `_data` folder

pages/views/_layouts/landing.njk
```diff
@@ -15,6 +15,8 @@
       {% endif %}
       {{ content | safe or 'No content.' }}
     </div>
+    {% from 'benefit.njk' import benefit with context %}
+    {{ benefit ( site.benefit ) }}
```
abarmina, line 19: mb can we immediately use {% for %} here (whole file benefit.njk)? without macro

pages/src/benefit-component/benefit-component.less
```diff
@@ -0,0 +1,65 @@
...
+}
+
+@media @lg {
+  .benefit-container {
+      justify-content: space-between;
+    }
+}
+
+@media @xl {
+  .benefit{
```
abarmina, line 55:
```suggestion
  .benefit {
```

ala-n (approved):

dshovchko (approved):

pages/src/benefit-component/benefit-component.less
```diff
@@ -0,0 +1,65 @@
+@import (reference) "../common/variables.less";
+
+.benefit {
+  color: #fff;
```
dshovchko, line 4: Just a question, isn't there a way to define colors through variables? After all, I've heard there will be light and dark themes. And the case of specifying a specific value in such a case will play a nasty trick.

abarmina: I think we can put it in this variable file `src/common/variables.less`

ala-n: Text color should be cascading, so we will need to have it defined on the parent container ideally.

PR #613, opened 2021-11-05
Fix style benefit and footer

Anna-MariiaPetryk:
# PR Details

- ignored filter created to skip 'All' item in the lists
- Benefits Component styles rework
- Cosmetic fixes


pages/11ty/sort.filter.js
```diff
@@ -8,6 +8,9 @@ module.exports = (config) => {
     return [...values].sort(comparer);
   };
 
+  /** filter by ignore */
+  const filterIgnore = (collection) => collection.filter( item => !item.data.ignore )
```
ala-n, line 12: it doesn't look to be a sort 11ty filter
it's better to create a separate module

pages/views/_includes/landing/benefits.njk
```diff
@@ -6,10 +6,12 @@
         <div class="benefit-icon">
           {% include "static/assets/" + item.icon %}
         </div>
-        <div class="benefit-content-name">{{ item.name }}</div>
-        <p class="benefit-content-text">
-          {{ item.text }}
-        </p>
+        <div>
```
ala-n, line 9: do we need an empty div?

PR #627, opened 2021-11-09
tech(redesign): sidebar imporvement

fshovchko:
# PR Details

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.


pages/src/navigation/sidebar/sidebar.less
```diff
@@ -23,6 +23,10 @@ esl-d-sidebar {
   }
 
   @media @xs-sm {
+    position: fixed;
+    top: 56px;
+    height: calc(100vh - 56px);
+
```
ala-n, line 29: Mobile settings is the default settings applied to an element, it produces code duplication in case we overwrite it in the media query

pages/src/common/helpers.less
```diff
@@ -108,3 +108,8 @@
   white-space: nowrap;
   border: 0;
 }
+
+// Icon link hover effect
+.icon-link:hover svg {
+  filter: drop-shadow(2px 2px 1px @link-fg);
```
ala-n, line 114:
```suggestion
  filter: drop-shadow(0px 0px 4px @link-fg);
```

pages/src/navigation/header/header.less
```diff
@@ -7,6 +7,8 @@
   background-color: @landing-bg-color;
 
   max-height: 56px;
+
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, .5);
```
ala-n, line 11: Unfortunately, we need to rethink this approach because right now shadow doesn't look good on the landing page with the marquee. We should not display shadow until we scroll the page.

We need to think about possible solutions. My current idea is to create a shadow extra element after the main section scroll so it can consume `at-start` marker. And display it with position fixed under the header to produce a shadow effect.

pages/src/navigation/sidebar/sidebar.ts
```diff
@@ -59,11 +60,13 @@ export class ESLDemoSidebar extends ESLToggleable {
   }
 
   protected onShow(params: SidebarActionParams) {
+    localStorage.setItem('sidebar-expanded', JSON.stringify(true));
```
ala-n, line 63: Can be simplified to storing single param 'sidebar-collapsed' that can be 'true' or undefined (removed)
There is no need an object or `JSON.parese`.
Moreover, we should have a sidebar opened by default for a first-time visits.

pages/src/navigation/sidebar/sidebar.ts
```diff
@@ -59,11 +60,13 @@ export class ESLDemoSidebar extends ESLToggleable {
...
     this._animation = !params.immediate;
     super.onShow(params);
     this.expandActive(params.initiator === 'init');
   }
   protected onHide(params: SidebarActionParams) {
+    localStorage.setItem('sidebar-expanded', JSON.stringify(false));
```
ala-n, line 69: Basically, it's better to move this logic to a separate single method `storeState` for both show and hide action.
And we shouldn't store state right before animation because the animation is a primary action that we should do.

pages/views/_includes/navigation/sidebar.njk
```diff
@@ -28,3 +28,14 @@
     <esl-trigger class="sidebar-trigger-btn sidebar-arrows-icon" target="#sidebar" mode="toggle"></esl-trigger>
   </nav>
 </esl-d-sidebar>
+<script>
+  if (!JSON.parse(localStorage.getItem('sidebar-expanded') || '{}'))
```
ala-n, line 32: simplify with updates mentioned above

pages/views/_includes/navigation/sidebar.njk
```diff
@@ -28,3 +28,14 @@
...
+  {
```
ala-n, line 33: should be on the same line with if condition

pages/views/_includes/navigation/sidebar.njk
```diff
@@ -28,3 +28,14 @@
...
+    const sidebar = document.querySelector('#sidebar');
+    sidebar.classList.remove('open');
+
+    sidebar.querySelector('.sidebar-trigger-btn').removeAttribute('active', '');
```
ala-n, line 37:
```suggestion
    sidebar.querySelector('.sidebar-trigger-btn').removeAttribute('active');
```

pages/views/_includes/navigation/sidebar.njk
```diff
@@ -28,3 +28,14 @@
...
+    const menus = sidebar.querySelectorAll('.sidebar-nav-secondary')
+      .forEach((menu) => menu.classList.remove('open'));
```
ala-n, line 39:
```suggestion
    Array.from(sidebar.querySelectorAll('.sidebar-nav-secondary')).forEach((menu) => menu.classList.remove('open'));
```

ala-n (changes requested):

pages/src/navigation/sidebar/sidebar.ts
```diff
@@ -37,9 +37,13 @@ export class ESLDemoSidebar extends ESLToggleable {
     ESLMediaQuery.for('@+MD').removeListener(this.onBreakpointChange);
   }
 
+  protected storeState() {
+    this.open ? localStorage.removeItem('sidebar-collapsed') : localStorage.setItem('sidebar-collapsed', 'true');
```
dshovchko, line 41: I guess it is enough to store any string except an empty string. You don't have to write 'true', you can just write '1'.

pages/src/page/landing-layout.less
```diff
@@ -1,6 +1,7 @@
 @import (reference) "../common/variables.less";
 @import "./base-layout.less";
 
+
```
dshovchko, line 4: extra row

pages/src/navigation/sidebar/sidebar.less
```diff
@@ -7,9 +7,9 @@ esl-d-sidebar {
   background: @nav-dark-bg;
 
   position: fixed;
+  top: 56px;
   z-index: 10;
-  left: 0;
-  height: 100vh;
+  height: calc(100vh - 56px);
```
dshovchko, line 12: For cases like this, it is better to define the value of 56px in a variable and use the variable in your code. This makes it easier to understand that your top position value and height value are related to each other.

dshovchko (approved):

abarmina (approved):

ala-n (approved):

PR #634, opened 2021-11-12
feat(esl-utils): array unwrap utility

fshovchko:
Function that accepts any value and returns first element of array-like value or returns it back if it's non array-like


src/modules/esl-utils/misc/array.ts
```diff
@@ -2,6 +2,21 @@ import {identity} from './functions';
 
 type Tuple<T> = [T?, T?];
 
+interface ArrayLike<T = any> {
+  [key: number]: T;
+  length: number;
+}
+
+const isArrayLike = (value: any): value is ArrayLike => {
```
ala-n, line 10: I think we can also export this utility to outside world

src/modules/esl-utils/misc/array.ts
```diff
@@ -20,6 +35,13 @@ export const wrap = <T>(arr: undefined | null | T | T[]): T[] => {
   return [arr];
 };
 
+/** Unwrap passed array
+ * @returns first array element or undefined if passed array is empty
+ */
```
ala-n, line 40:
```suggestion
/** Unwraps and returns the first element if passed object is array-like, returns original object otherwice */
```

src/modules/esl-utils/misc/test/array.test.ts
```diff
@@ -48,6 +48,75 @@ describe('misc/array helper tests', () => {
...
 type V = string | number;
 
+describe('unwrap', () => {
+  document.body.innerHTML = '<div>1</div><div>2</div>';
+
+  test('NodeList to unwrap into first Node', () => {
+    expect(unwrap(document.querySelectorAll('div'))).toEqual(document.querySelector('div'));
+  });
+
+  test.each([
```
ala-n, line 58: Could you please reformat test cases to a single lines to improve readability of the test

src/modules/esl-utils/misc/array.ts
dshovchko: That's a great idea!
But I think a more appropriate location would be an object so it is still a check that the object is a pseudo-array.
Also, I think we need to check that the object is iterable, i.e. we need a method to check - isIterable() and call it inside method unwrap(). Although I'm not sure about it. :confused:

ala-n: Great addition.
I'm ok with iterables in case we still will be compatible with IE, but it is a low priority.
Also regarding our current main case, I think we should not unwrap string.
Basically our real cases for now:
 - array or element - so situation, when we are not sure, is the input will be an array or item, with the assumption that the item will not be array-like
 - NodeList or Node - so we unwrap NodeList but not a node 
 - JQuery collection or HTML Element - again we are not sure about the input but want to have the JQuery collection unwrapped (we can not include even types of JQ to the library for sure, but it will be nice to have that case covered)

dshovchko (approved):

yadamskaya (approved):

julia-murashko (approved):

src/modules/esl-utils/misc/array.ts
```diff
@@ -13,13 +14,22 @@ export const tuple = <T>(arr: T[]): Tuple<T>[] => arr.reduce((acc: Tuple<T>[], e
...
+/** Unwraps and returns the first element if passed object is array-like, returns original object otherwise */
+export function unwrap(value: []): undefined;
+export function unwrap<T>(value: (ArrayLike<T> & {0: T}) | T): T;
+export function unwrap<T extends Node>(value: NodeListOf<T>): T | undefined;
+export function unwrap(value: any): any;
+export function unwrap(value: any): any {
+  return isArrayLike(value) ? value[0] : value;
+}
+
 /** Make array values unique */
```
julia-murashko, line 33:
```suggestion
/** Makes array values unique */
```

src/modules/esl-utils/misc/object.ts
```diff
@@ -1,10 +1,31 @@
-export const isObject = (obj: any): obj is Record<string, any> => obj && typeof obj === 'object';
+/** Checks that passed value is object, but not a callable-object (function) */
+export const isObject = (obj: any): obj is Record<string, any> => !!obj && typeof obj === 'object';
+/** Check that passed value is an object or function */
```
julia-murashko, line 3:
```suggestion
/** Checks that passed value is an object or function */
```

src/modules/esl-utils/misc/object.ts
```diff
@@ -1,10 +1,31 @@
...
-
+/** Checks if the passed value is primitive */
+export const isPrimitive = (obj: any): obj is string | number | boolean | symbol | undefined | null =>
+  obj === null ||
+  typeof obj === 'undefined' ||
+  typeof obj === 'string' ||
+  typeof obj === 'number' ||
+  typeof obj === 'boolean' ||
+  typeof obj === 'symbol';
+/** Check that passed object is prototype of some class */
```
julia-murashko, line 13:
```suggestion
/** Checks that passed object is prototype of some class */
```

julia-murashko (approved):

Issue #522, opened 2021-10-06
[🚀for esl-utils]: extend ESLTraversingQuery with ability to restrict global scope


ala-n:
As an ESLTraversingQuery consumer, I want to have complete control over the search scope.

In the current state simple selector or selector base for the ESLTraversingQuery is assumed as a query to search in whole document e.g.:
`ESLTraversingQuery.first('.class') === document.querySelector('.class')`
`ESLTraversingQuery.first('#simple .selector') === document.querySelector('.selector')`
`ESLTraversingQuery.first('.selector::next(div)') === ESLTraversingQuery.first('::next(div)', document.querySelector('.selector'));`

See [current ESLTraversingQuery behavior](https://github.com/exadel-inc/esl/blob/HEAD/src/modules/esl-traversing-query/README.md) for details.

**Requested** to add optional scope third param to define the scope to find the global part of the query.
The scope should be equal document by default.

In addition, extend the base element with the ability to pass null in addition to undefined (with the same meaning). (To simplify cases when base element came from DOM API functions where indicates unexisting result)

Note: extend unit test accordingly.

**API**
- ` ESLTraversingQuery.first(query: string, base?: Element | null, scope: Element | Document = document): Element | null; `
- ` ESLTraversingQuery.all(query: string, base?: Element | null, scope: Element | Document = document): Element[] `

**Test Cases | examples**
Context: 
```html
<section class="container"> 
     <div id="row1" class="row" data-test="1">
        <button id="btn1" class="btn btn-1"></button>
        <button id="btn2" class="btn btn-2" data-test="prev"></button>
        <button id="btn3" class="btn btn-3"></button>
        <button id="btn4" class="btn btn-4"></button>
        <button id="btn5" class="btn btn-5" data-test="next"></button>

        <article class="col-1">Hello!</article>
      </div>
      <div id="row2" class="row">
        <article class="col-2">Hello 2!</article>
        <button id="btn6" class="btn btn-6" data-test="next"></button>
      </div>
</section>
```

```ts
ESLTraversingQuery.first('#row1', undefined) // <div id="row1" ...
ESLTraversingQuery.first('#row1', null) // <div id="row1" ...
ESLTraversingQuery.first('#row1', null, $section) // <div id="row1" ...
ESLTraversingQuery.first('#row1', null, $row2) // null
```

PR #639, opened 2021-11-15
feat(esl-traversing-query): add ability to restrict global scope

Anna-MariiaPetryk:
Closes: #522

### DIFF
--- a/src/modules/esl-traversing-query/core/esl-traversing-query.ts
+++ b/src/modules/esl-traversing-query/core/esl-traversing-query.ts
@@ -94,21 +94,22 @@ export class TraversingQuery {
     return uniq(result);
   }
 
-  static traverse(query: string, findFirst: boolean, base?: Element) {
+  static traverse(query: string, findFirst: boolean, base?: Element | null, scope: Element | null | Document = document) {
+    if (scope === null) throw new Error();
     const parts = query.split(this.PROCESSORS_REGEX).map((term) => term.trim());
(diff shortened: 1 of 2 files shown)


src/modules/esl-traversing-query/core/esl-traversing-query.ts
```diff
@@ -94,21 +94,22 @@ export class TraversingQuery {
     return uniq(result);
   }
 
-  static traverse(query: string, findFirst: boolean, base?: Element) {
+  static traverse(query: string, findFirst: boolean, base?: Element | null, scope: Element | null | Document = document) {
+    if (scope === null) throw new Error();
```
ala-n, line 98: In case we throw an error - there is no reason to allow null value at all

ala-n: I'd recommend restricting type and getting rid of an error (the user will have it anyway if he passed a null)
From an API point of view, we are also do not allow undefined or null scope, it is the required field for TraversingQuery
`base` element accepts null just for the purpose of being searched using query without base, e.g. queries like `#id::next` and etc.

ala-n (changes requested):

ala-n (approved):

yadamskaya (approved):

julia-murashko (approved):

dshovchko (approved):

PR #667, opened 2021-11-26
feat(esl-footnotes): add back-to-note-label attribute and renderIndex for connected note

dshovchko:
In the scope of PR:
- added separate method to render the content of the connected note
- added back-to-note-label attribute to outputting title for the button


src/modules/esl-footnotes/core/esl-note.ts
```diff
@@ -46,9 +46,17 @@ export class ESLNote extends ESLBaseElement {
...
     return this._index;
   }
+  public set index(value: number) {
+    this._index = value;
+    this.innerHTML = this.renderedIndex;
+  }
+
+  public get renderedIndex(): string {
+    return `${this._index}`;
+  }
```
ala-n, line 59: Is it in purpose to be rewritten?

ala-n (approved):

src/modules/esl-footnotes/core/esl-note.ts
dshovchko: That's right.

dshovchko: And since the rewritten value must be used by the footnote component.

fshovchko (approved):

PR #689, opened 2021-12-01
feat(esl-utils): `ESLVSizeCSSProxy` utility to fix 100vh problem is introduced

ala-n:
`ESLVSizeCSSProxy` is observing window resizes and creates custom CSS variables with exact
viewport size (independently of Android / iOS browser navigation toolbar)


src/modules/esl-utils/fixes/viewport.ts
```diff
@@ -0,0 +1,42 @@
+import {rafDecorator} from '../async/raf';
+import {DeviceDetector} from '../environment/device-detector';
+
+/**
+ * Small utility to provide 100vw and 100vh alternative CSS Variables
+ * - vwProp (default: `--100vw`) - is a scroll independent viewport width value
+ * - vhProp (default: `--100vh`) - is a device independent viewport height value (aso known as 100vh iOS fix)
```
nattallius, line 7:
```suggestion
 * - vhProp (default: `--100vh`) - is a device independent viewport height value (also known as 100vh iOS fix)
```

nattallius (approved):

dshovchko (approved):

Issue #674, opened 2021-11-28
[🚀esl-core]: extend library output with SASS styles representation


ala-n:
As an ESL architect, I want to have more popular CSS preprocessor syntaxes to be presented in the library output.

Request for adding an SCSS syntax to the library output.

Assumptions:
- SCSS will be generated automatically for library `modules` output. 
- LESS syntax still will be a source for ESL `modules`
- LESS features can be limited in purpose to have stable and valid convertation

**Priority**: Normal

**PR** #694

ala-n:
According to the followup discussions, the proposal is canceled.
ESL going to deliver only pure CSS output starting from 6th version of library

PR #694, opened 2021-12-04
feat: extend library output with scss

NastaLeo:
https://github.com/exadel-inc/esl/issues/674

Installed less to sass converter that add SCSS syntax to the library output


ala-n:
Draft the PR before style distribution strategy discussion

ala-n:
Canceled

PR #701, opened 2021-12-06
fix(esl-footnotes): fix inappropriate behaviour

dshovchko:
In the scope of this PR:
 - fixed scrollIntoView() - it became more responsive and also supports horizontal scrolling
 - fixes in the bundle of three components ESLNote, ESLFootnote, ESLTooltip. There was a problem with inappropriate behaviour when activating a note from footnotes using the return-to-note button.


yadamskaya (approved):

abarmina (approved):

src/modules/esl-tooltip/core/esl-tooltip.ts
```diff
@@ -41,8 +41,11 @@ export class ESLTooltip extends ESLPopup {
     return els.length ? els[els.length - 1] : null;
   }
 
+  public static isOpen(): boolean {
+    return ESLTooltip.sharedInstance.open;
+  }
```
ala-n, line 46: In case we gonna use in such API representation - please use getter accessor to be consistent with API of other components
```suggestion
  public static get open(): boolean {
    return ESLTooltip.sharedInstance.open;
  }
```

ala-n (approved):

NastaLeo (approved):

src/modules/esl-tooltip/core/esl-tooltip.ts
dshovchko: Thanks! Fixed

PR #707, opened 2021-12-07
feat(esl-popup): couple new features and refactoring

dshovchko:
In scope of this PR:
 - added `disable-activator-observation` attribute to let popup leaves viewport without hiding
 - added new `fit-on-major` value for the behavior attribute so that the popup update its position in order to fit on the major axis only
 - added new `fit-on-minor` value for the behavior attribute so that the popup update its position in order to fit on the minor axis only
 - added RTL support, resulting in a popup mirrored relative to the major axis
 - updated TSX shapes
 - refactor code to achieve the best performance during popup position updating

### DIFF
--- a/src/modules/esl-popup/core/esl-popup.ts
+++ b/src/modules/esl-popup/core/esl-popup.ts
@@ -2,6 +2,7 @@ import {range} from '../../esl-utils/misc/array';
 import {ExportNs} from '../../esl-utils/environment/export-ns';
 import {attr, boolAttr, jsonAttr} from '../../esl-base-element/core';
 import {bind} from '../../esl-utils/decorators/bind';
+import {memoize} from '../../esl-utils/decorators/memoize';
 import {ready} from '../../esl-utils/decorators/ready';
 import {prop} from '../../esl-utils/decorators/prop';
 import {rafDecorator} from '../../esl-utils/async/raf';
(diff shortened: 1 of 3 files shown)


src/modules/esl-popup/core/esl-popup.ts
```diff
@@ -91,32 +97,36 @@ export class ESLPopup extends ESLToggleable {
...
   }
 
-  /** Is position along vertical axis? */
-  protected get _isPositioningAlongVertical() {
-    return ['top', 'bottom'].includes(this.position);
+  /** Checks that the position along the vertical axis */
+  @memoize()
+  protected get _isMajorAxisVertical(): boolean {
+    return !isMajorAxisHorizontal(this.position);
   }
```
ala-n, line 115: Cosmetic: it looks like the memoize will be much heavier here then exact check inside the `isMajorAxisHorizontal`

ala-n (approved):

src/modules/esl-popup/core/esl-popup.ts
dshovchko: I was of the same opinion at first.
But I wondered to check and understand with the help of benchmarks. I don't seem to have the same understanding now. 
We can discuss it.

Issue #544, opened 2021-10-10
[🚀gh-pages]: GSS integration


ala-n:
As an ESL demo site holder, I want to have search functionality supported on the https://exadel-inc.github.io/esl

It is required to process the following steps:
- [x] investigate site search providers that can be integrated with the static site hosted on GitHub pages
- [x] investigate exact steps to use and integrate Google Site Search (as the most probable candidate)
- [x] create an account (if needed) in the chosen search provider 
*IMPORTANT NOTE*: a core team member should be the account owner, please contact @exadel-inc/esl-core-team to create an account
- [x] integrate the chosen site search provider with the current https://exadel-inc.github.io/esl site

Note:
 - Provider should be stable, free, and compliant with the usage in Open Source products.
 - Provider should support UI customizations
 
**Priority**: Normal

PR #723, opened 2021-12-15
feat(gh-pages): GSS integration

abarmina:
# PR Details
Added support for site search using Google Site Search:

- account created;
- created a new demo Search Box component;
- rewritten the default styles of the Google search engine;
- added animation to the search in the header.

@ala-n is the owner of the Google account.

## Related Issue

https://github.com/exadel-inc/esl/issues/544


Anna-MariiaPetryk (changes requested):
when the results are more than one page and I go to the next page I see "No Results"
in general very cool

abarmina:
hi @Anna-MariiaPetryk!
as I understand this is the logic of the PSE (Programmable Search Engine):

> The result number in PSE that shows up initially is just an estimated number. This estimate may not be accurate. This is done to serve the results faster. As you navigate to the next results set you will start seeing the exact result number. 
> 
> PSE will only serve 10 pages (100 results). Since the estimated results are more than 100 you see the results navigation link 1 2 3 4 5 6 7 8 9 10 at the bottom of the 1st results set. As you navigate to next results set i.e by clicking on pagination 2 3 etc you will start seeing the exact result number. Now, if you directly click on 9 or 10 links or the last result page link, it may so happen that there are no results and therefore you get the message "no results found". This is how PSE works.
>
>*[link to this answer](https://support.google.com/programmable-search/thread/114368882/initial-search-shows-172-million-results-click-to-next-page-and-there-are-two-results-what-gives?hl=en)

I propose to enable `Results Browsing History` option, it helps to jump back to the previous search results page. 
@ala-n, what do you think?

search

pages/views/_includes/navigation/footer.njk
```diff
@@ -20,8 +20,8 @@
   <div class="footer-common-links-container">
     <div class="footer-common-links">
       <div class="footer-copyright">
-        ESL v{{ env.version }} © 2021 <a class="footer-link" href="https://exadel.com">Exadel</a>, Inc.
-        Code licenced under <a class="footer-link" href="https://github.com/exadel-inc/esl/blob/main/LICENSE">MIT</a>
+        ESL v{{ env.version }} © 2021 <a class="footer-link" href="https://exadel.com" target="_blank">Exadel</a>, Inc.
```
dshovchko, line 23: Can we change the year automatically instead of manually?

dshovchko (approved):

pages/views/_includes/navigation/footer.njk
ala-n: If I remember correctly copyright year is attached to the license and should not change when 2022 begins.

ala-n:
> Results Browsing History

Hey @abarmina, @Anna-MariiaPetryk 
Results Browsing History option is already enabled in our PSE settings.
Give me a little time to investigate what is going on and how can we affect UX in that case.

pages/views/_includes/navigation/footer.njk
abarmina: yes, we can add it as a value and use on our page (as `version` in this case)

ala-n (approved):

NastaLeo (approved):

PR #741, opened 2021-12-26
bugfix: Tests unhandled rejection fix 

fshovchko:
In the scope of this PR i fixed issue, where test was throwing an unhandled promise rejection

### DIFF
--- a/src/modules/esl-utils/async/test/debounce.test.ts
+++ b/src/modules/esl-utils/async/test/debounce.test.ts
@@ -44,7 +44,9 @@ describe('async/debounce', () => {
     const promise = debounced.promise;
     debounced.cancel();
     return promise.then(
-      () => expect(fn).toBeCalledTimes(-1),
+      () => {
+        throw new Error('Promise shouldn`t be resolved');
+      },
       () => expect(fn).toBeCalledTimes(0)
     );
   });
--- a/src/modules/esl-utils/async/test/debounce.test.ts
+++ b/src/modules/esl-utils/async/test/debounce.test.ts
@@ -40,12 +40,12 @@ describe('async/debounce', () => {
   test('cancel debounce', () => {
     const fn = jest.fn();
     const debounced = debounce(fn, 10);
-    const promise = debounced.promise;
     debounced();
-    debounced.cancel();
-
+    setTimeout(() => debounced.cancel());
+    const promise = debounced.promise;
+    jest.runAllTimers();
     return promise.catch(() => expect(fn).toBeCalledTimes(0));
-  }, 100);
+  });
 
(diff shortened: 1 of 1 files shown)


src/modules/esl-utils/async/test/debounce.test.ts
```diff
@@ -40,12 +40,12 @@ describe('async/debounce', () => {
   test('cancel debounce', () => {
     const fn = jest.fn();
     const debounced = debounce(fn, 10);
-    const promise = debounced.promise;
```
ala-n, line 43: That was the actual reason for an exception.
The promise should be extracted after the first call.

src/modules/esl-utils/async/test/debounce.test.ts
```diff
@@ -40,12 +40,12 @@ describe('async/debounce', () => {
...
     debounced();
-    debounced.cancel();
-
+    setTimeout(() => debounced.cancel());
+    const promise = debounced.promise;
+    jest.runAllTimers();
     return promise.catch(() => expect(fn).toBeCalledTimes(0));
```
ala-n, line 47:
```suggestion
    return promise.then(
      () => expect(fn).toBeCalledTimes(-1),
      () => expect(fn).toBeCalledTimes(0)
    );
```
That the second cause of the problems in that test, the final condition was false positive and doesn't catch incorrect handling.

src/modules/esl-utils/async/test/debounce.test.ts
```diff
@@ -40,12 +40,12 @@ describe('async/debounce', () => {
   test('cancel debounce', () => {
     const fn = jest.fn();
     const debounced = debounce(fn, 10);
-    const promise = debounced.promise;
     debounced();
-    debounced.cancel();
-
+    setTimeout(() => debounced.cancel());
+    const promise = debounced.promise;
```
ala-n, line 45: That's incorrect: 
`debounced.cancel()` should be called sync just after the first call
In case we call it in bounds of a separate task - that's a different type of test case.

```suggestion
    const promise = debounced.promise;
    debounced.cancel();
```

ala-n (changes requested):

dshovchko:
As for me, the functionality of the tests is restored if lines 43 and 44 swap places. 
The thing is that as far as I see from the code getting promise before calling delayed function leads to returning promise in the resolved state.

`
Object.defineProperty(debouncedSubject, 'promise', {
    get: () => deferred ? deferred.promise : Promise.resolve()
});
`

And variable `differed` will be defined when `debouncedSubject()` was called at least one time.

dshovchko (approved):

ala-n (approved):

Issue #748, opened 2021-12-30
[🔨 code style]: Placeholder Discussion 2022/01  


ala-n:
- [] Use explicit return types for TS
- [] ...

PR #762, opened 2022-01-10
[🔨 code style]: explicit return type (lint rule and fix)

ala-n:
Part of #748


src/modules/esl-media/core/esl-media.ts
```diff
@@ -374,8 +374,9 @@ export class ESLMedia extends ESLBaseElement {
     (this._provider) && this._provider.safeSeekTo(time);
   }
 
+  // TODO: simplify - ESLMediaQuery is lazy for primitives, it make sense to init it unconditionally
```
ala-n, line 377: ESLMediaQuery 2.0.0 has lightweight constant queries and ootb cache. So it's better to have 'ALL' or empty string default for query with assumption that the query is always available instead of managing everything manually.

dshovchko (approved):

yadamskaya (approved):

fshovchko (approved):

abarmina (approved):

julia-murashko (approved):

NastaLeo (approved):

PR #766, opened 2022-01-11
chore: add webpack bundle analyzer

fshovchko:
In the scope of this PR I added script that generates webpack stats.json file and then it's parsed with the help of webpack bundle analyzer plugin


package.json
```diff
@@ -53,6 +53,7 @@
     "build:less:cpy": "copyfiles \"src/**/*.less\" \".\" --up 1 --all",
     "build:less:css": "foreach -x \"lessc #{path} #{dir}/#{name}.css\" -i \"**/*.mixin.less\" -g \"modules/{*,*/*,{draft,esl-utils,esl-forms}/*/*}.less\"",
     "build:docs": "copyfiles \"src/**/*.md\" \".\" --up 1 --all",
+    "build:analyzer": "cd pages && webpack --json ../.report/stats.json && webpack-bundle-analyzer ../.report/stats.json -m static -r ../.report/stats.html",
```
ala-n, line 56: It's better to move it under the build-pages category as it does not relate to the main build.

ala-n (changes requested):
The reason for the request to add a bundle analyzer is to have an assumption of how the library will work on the real users' side.
There is a couple of problems with that right now:
 - webpack builds library in the dev mode instead of production
 - there is no min and (important) gzipped sizes in the report right now

ala-n (approved):

PR #779, opened 2022-01-18
chore(gh-pages): gss lazy loading

fshovchko:
In the scope of this PR I changed the way google search script loads. Now it's loading when pressed search icon on landing page header and on the search page


pages/views/_layouts/base.njk
```diff
@@ -22,7 +22,28 @@
...
+        });
+      }
+
+      function addScript() {
+        var gssScript = document.createElement('script');
+        gssScript.setAttribute('src', 'https://cse.google.com/cse.js?cx=3171f866738b34f02');
+        document.querySelector('head').appendChild(gssScript);
+      }
+    </script>
   {% endblock %}
```
ala-n, line 47: Should be part of the search component of the header.
We have client scripts and exact components to handle that.

pages/views/_layouts/base.njk
```diff
@@ -22,7 +22,28 @@
...
+          var searchBar = document.querySelector('.header-search-box');
+          var searchBarShowHandler = function() {
+            addScript();
+            searchBar.removeEventListener('esl:show', searchBarShowHandler)
+          }
+          searchBar.addEventListener('esl:show', searchBarShowHandler);
+        });
+      }
+
+      function addScript() {
```
ala-n, line 41: When it's appeared in the right place (our ts sources for pages) the are all helpers for loading will be presented

pages/views/_layouts/base.njk
```diff
@@ -22,7 +22,28 @@
     <link rel="stylesheet" type="text/css" href="{{ '/bundles/localdev.css' | url }}" />
   {% endblock %}
   {% block head %}
-    <script async src="https://cse.google.com/cse.js?cx=3171f866738b34f02"></script>
+    <script async>
```
ala-n, line 25: Just a small note: the inline script can not be async.
`async` is a first of all attribute for load process adjustments, inline script doesn't load it is already on the page.

ala-n (changes requested):
For the search result page, I'd recommend having some additional marker (like custom component wrapper or class) on the page to check and load `gss`.

ala-n (changes requested):
Does not accept from a UX point of view.
The first search request is leeds to glitch jump.
We should handle it using some spinners, etc so the visual result should be smooth independently of the technical way of loading.

pages/src/search/search.ts
```diff
@@ -0,0 +1,23 @@
...
+  static is = 'esl-d-search-page-wrapper';
+
+  @ready
+  protected connectedCallback(): void {
+    loadSearchScript();
+
+    (window as any).__gcse = {
+      parsetags: 'onload',
+      initializationCallback: (): void => this.afterSearchScriptLoad()
+    };
```
dshovchko, line 16: Cosmetic: for better understanding and clarity of code, it is better to put this code in a separate private method, such as initSearchScript()

dshovchko (approved):

pages/src/localdev.ts
```diff
@@ -51,13 +51,16 @@ import './esl-media-demo/test-media-source';
...
 import {ESLDemoSearchBox} from './navigation/header/header-search';
+import {ESLDemoSearchPageWrapper} from './search/search';
 
 ESLVSizeCSSProxy.observe();
 
 // Register Demo components
 ESLDemoSidebar.register();
 ESLDemoMarquee.register();
 ESLDemoSearchBox.register();
+ESLDemoSearchBox.register();
```
NastaLeo, line 62: Typo: double ESLDemoSearchBox.register();

pages/src/navigation/header/header-search.ts
```diff
@@ -19,13 +20,39 @@ export class ESLDemoSearchBox extends ESLToggleable {
...
+    };
+  }
+
+  private showSearchElements(params: ToggleableActionParams): void {
     afterNextRender(() => super.onShow(params));
     if (this.autofocus) {
       const $focusEl = TraversingQuery.first(this.firstFocusable, this) as HTMLElement;
       $focusEl && window.setTimeout(() => $focusEl.focus(), parseNumber(this.postClsDelay));
     }
+    const loadingAnimationEL = this.querySelector('.animation-loading')!;
```
NastaLeo, line 54: Maybe I missed up something, what means '!' at the end of sentence?

NastaLeo (approved):

pages/src/navigation/header/header-search.ts
ala-n: `!` at the end of the object marks it any for typescript

pages/src/search/search.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  private initSearchScript(): void {
+    (window as any).__gcse = {
+      parsetags: 'onload',
+      initializationCallback: (): void => this.afterSearchScriptLoad()
+    };
+  }
+
+  private afterSearchScriptLoad(): void {
+    const loadingAnimationEL = this.querySelector('.animation-loading')!;
```
ala-n, line 23: Please remove strict type disabling.
The `Element.[proto].querySelector` returns Element | null.
CSSClassUtils.add expect Element (not nullable).
Typescript says "here is a "null pointer"-like problem, please make your script safe",
Do not ignore that, better fix it.

pages/src/navigation/header/header-search.ts
ala-n: But the same as in the point below, it's better properly check the null result instead of disabling types

pages/src/navigation/header/header-search.ts
```diff
@@ -19,13 +20,39 @@ export class ESLDemoSearchBox extends ESLToggleable {
...
+      });
+    }
+  }
+
+  private initSearchScript(params: ToggleableActionParams): void {
+    (window as any).__gcse = {
+      parsetags: 'onload',
+      initializationCallback: (): void => this.showSearchElements(params)
+    };
+  }
```
ala-n, line 46: DRY

ala-n (changes requested):
We have the `pages/src/search/search-script.ts` - that the place that should deliver a safe API for load and controlling Google script.

1) You have a `loadScript` it's safe enough to load the script and make it once. 
But it doesn't check the real state of the google script - that's ok.
2) So you have 
```
(window as any).__gcse = {
      parsetags: 'onload',
      initializationCallback: (): void => /** GSS ready*/
};
```
Place this logic to the `pages/src/search/search-script.ts`.
Then promicify it - Promise is a well native API for something async.
3) Create a chain `loadScript ` then `awaitGSSReady` and memorize the result.
Super you have a simple API that allows you to request GSS, in one line of code and be sure that it loaded once, and correctly handles GSS behavior.

ala-n (approved):

dshovchko (approved):

NastaLeo (approved):

PR #788, opened 2022-01-20
feat: esl-popup-placeholder and moving to document.body

dshovchko:
In the scope of this PR:
 - created esl-toggleable-placeholder element as a base placeholder for all toggleable derivatives
 - added the ability for the target to get origin from the placeholder element on esl-trigger
 - created esl-popup-placeholder element
 - added the ability to esl-popup to move into body in case when it is not in body
 - reworked esl-popup examples


pages/views/examples/popup-transition.njk
```diff
@@ -73,6 +73,8 @@ aside:
   </div>
 </section>
 
+<hr>
```
ala-n, line 76:
```suggestion
<hr/>
```

pages/views/examples/popup-transition.njk
```diff
@@ -91,6 +93,8 @@ aside:
...
```
ala-n, line 96:
```suggestion
<hr/>
```

pages/views/examples/popup-transition.njk
```diff
@@ -109,6 +113,8 @@ aside:
...
```
ala-n, line 116:
```suggestion
<hr/>
```

pages/views/examples/popup-transition.njk
```diff
@@ -127,6 +133,8 @@ aside:
...
```
ala-n, line 136:
```suggestion
<hr/>
```

pages/views/examples/popup-transition.njk
```diff
@@ -145,6 +153,8 @@ aside:
...
```
ala-n, line 156:
```suggestion
<hr/>
```

pages/views/examples/popup.njk
```diff
@@ -60,6 +62,8 @@ aside:
   <esl-trigger target="#popup2" mode="toggle" class="btn btn-sec-blue">Bottom</esl-trigger>
 </section>
 
+<hr>
```
ala-n, line 65:
```suggestion
<hr/>
```

pages/views/examples/popup.njk
```diff
@@ -69,12 +73,66 @@ aside:
...
+  <esl-trigger target="#popup04" mode="toggle" class="btn btn-sec-blue">Right</esl-trigger>
 </section>
 
 <section class="row justify-content-between">
   <esl-trigger target="#popup02" mode="toggle" class="btn btn-sec-blue">Bottom</esl-trigger>
   <esl-trigger target="#popup02" mode="toggle" class="btn btn-sec-blue">Bottom</esl-trigger>
   <esl-trigger target="#popup02" mode="toggle" class="btn btn-sec-blue">Bottom</esl-trigger>
 </section>
+
+<hr>
```
ala-n, line 88:
```suggestion
<hr/>
```

pages/views/examples/popup.njk
```diff
@@ -69,12 +73,66 @@ aside:
...
+    <div style="height: 250px; width:50px;"></div>
+    <span class="esl-popup-arrow"></span>
+  </esl-popup>
+  <esl-popup class="demo-right-popup" position="right">
+    <div style="height: 250px; width:50px;"></div>
+    <span class="esl-popup-arrow"></span>
+  </esl-popup>
+</section>
+
+<hr>
```
ala-n, line 116:
```suggestion
<hr/>
```

ala-n (commented):
Looks good but have a couple of points to discuss:
1) Is the fact that we copy all attributes right, the proxy cannot use them
2) Another point that we missed in the previous discussions is that now it is not technically a truthful proxy...

dshovchko:
> Looks good but have a couple of points to discuss:
> 
> 1. Is the fact that we copy all attributes right, the proxy cannot use them
> 2. Another point that we missed in the previous discussions is that now it is not technically a truthful proxy...

Let's discuss this on Monday.

fshovchko (approved):

fshovchko (approved):

src/modules/esl-popup/core/esl-popup.ts
```diff
@@ -121,6 +124,21 @@ export class ESLPopup extends ESLToggleable {
...
+  /** Moves popup into document.body */
+  protected moveToBody(): void {
+    const {parentNode, $placeholder} = this;
+    if (parentNode && parentNode !== document.body) {
+      if ($placeholder) {
+        // to be safe and prevent leaks
+        $placeholder.parentNode?.removeChild($placeholder);
+      }
+      // replace this with placeholder element
+      this.$placeholder = ESLPopupPlaceholder.from(this);
```
ala-n, line 136: Cosmetic: do we need to recreate elements? In case we going to have only a single instance per popup it makes sense to create a lazy getter. That simplifies our work with DOM because then 
`parentNode.replaceChild(this.$placeholder, this);`
Will work perfectly independently of preconditions.

Point to discuss: behavior on popup disconnection ... (keeping in mind that move also producing disconnection)

src/modules/esl-toggleable/core/esl-toggleable-placeholder.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (!this.$origin) return;
+
+    [...this.$origin.attributes]
+      .filter((attr) => (this.constructor as typeof ESLToggleablePlaceholder).allowedAttrs.includes(attr.nodeName))
+      .forEach((attr) => {
+        const {nodeName, nodeValue} = attr;
+        if (nodeValue) {
+          this.setAttribute(nodeName === 'id' ? 'original-id' : nodeName, nodeValue);
+        }
+      });
```
ala-n, line 38:
```suggestion
    (this.constructor as typeof ESLToggleablePlaceholder).allowedAttrs.forEach((name) => {
        const value = this.$origin.getAttribute(name);
        const altName = name=== 'id' ? 'original-id' : name; // Or better move it out of generic logic
        this.setAttribute(altName, value); // No checks or constraints required
    });
```

src/modules/esl-popup/core/esl-popup.ts
dshovchko: Created the [issue](https://github.com/exadel-inc/esl/issues/795) to investigate behavior and solve some unusual cases.

src/modules/esl-toggleable/core/esl-toggleable-placeholder.ts
dshovchko: Thanks! Done

ala-n (approved):

src/modules/esl-popup/core/esl-popup.ts
ala-n: Thanks!

abarmina (approved):

src/modules/esl-popup/core/esl-popup.ts
```diff
@@ -121,6 +124,21 @@ export class ESLPopup extends ESLToggleable {
     return RTLUtils.isRtl(this) ? 1 - ratio : ratio;
   }
 
+  /** Moves popup into document.body */
+  protected moveToBody(): void {
+    const {parentNode, $placeholder} = this;
+    if (parentNode && parentNode !== document.body) {
+      if ($placeholder) {
+        // to be safe and prevent leaks
+        $placeholder.parentNode?.removeChild($placeholder);
```
julia-murashko, line 133: Minor and only suggestion: Mb we can simplify 
```suggestion
if (!parentNode || parentNode === document.body) return;

// to be safe and prevent leaks
$placeholder && $placeholder.parentNode?.removeChild($placeholder);
```

julia-murashko (approved):

src/modules/esl-popup/core/esl-popup.ts
dshovchko: Cool! Thanks!

PR #794, opened 2022-01-25
feat(esl-popup): container attribute to set up bounds of visibility

dshovchko:
In scope of this PR:
- added container attribute to set up bounds of visibility (esl-popup, esl-note, esl-note-group)
- refactored code to avoid duplication
- updated expamples


src/modules/esl-footnotes/core/esl-note-group.ts
```diff
@@ -4,17 +4,25 @@ import {ready} from '../../esl-utils/decorators/ready';
 
 import type {ESLNote} from './esl-note';
 
+interface Attribute {
+  name: string;
+  value: string;
+}
```
ala-n, line 10: Let's discuss changes in esl-note-group.ts today
Attribute propagation logic looks a little bit overcomplicated.

ala-n (commented):
Have no concerns except esl-note-group.ts changes

src/modules/esl-footnotes/core/esl-note-group.ts
dshovchko: You are right! It looks like an elephant on the bike. :)
I have already reworked esl-note-group.
What do you think about the new variant?

ala-n (approved):

abarmina (approved):

fshovchko (approved):

NastaLeo (approved):

PR #809, opened 2022-02-01
Footnotes fixes

dshovchko:
- fixed esl-note-group attributes propagation
- fixed behavior during connect/disconnect callbacks


fshovchko (approved):

src/modules/esl-footnotes/core/esl-note.ts
```diff
@@ -173,9 +175,7 @@ export class ESLNote extends ESLBaseElement {
   protected restore(): void {
     this.linked = false;
     this._$footnotes = null;
-    if (this.html) {
-      this.innerHTML = this.html;
-    }
```
ala-n, line 178: Why did we refuse to revert inner content on unlink?

ala-n (approved):

src/modules/esl-footnotes/core/esl-note.ts
dshovchko: Instead of reverting content, we set the index to 0.
After this update will be the next rules for note:
 - in not linked stated it will be 0;
 - in linked but with ignore-footnotes declaration it will be * (or will defined by standalone-label attribute);
 - in an ordinary linked state it will be the index in order from footnotes.

From real user cases, it figures out that reverting content leads to blinking content during resizing in the case note content is huge (or maybe just large).

NastaLeo (approved):

PR #810, opened 2022-02-01
Popup: add containerEl param to action params

dshovchko:
Added param to pass container element to popup

### DIFF
--- a/src/modules/esl-popup/core/esl-popup.ts
+++ b/src/modules/esl-popup/core/esl-popup.ts
@@ -113,6 +113,14 @@ export class ESLPopup extends ESLToggleable {
     return this.container ? TraversingQuery.first(this.container, this) as HTMLElement : this._containerEl;
   }
 
+  /** Get the size and position of the container */
+  protected get containerRect(): Rect {
+    const {$container} = this;
+    if (!$container) return getWindowRect();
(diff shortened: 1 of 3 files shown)


fshovchko (approved):

src/modules/esl-popup/core/esl-popup.ts
```diff
@@ -106,8 +109,16 @@ export class ESLPopup extends ESLToggleable {
 
   /** Container element that define bounds of popups visibility */
   @memoize()
-  protected get $container(): HTMLElement | null {
-    return this.container ? TraversingQuery.first(this.container, this) as HTMLElement : null;
+  protected get $container(): HTMLElement | undefined {
+    return this.container ? TraversingQuery.first(this.container, this) as HTMLElement : this._containerEl;
+  }
```
ala-n, line 114: Is that right that the `$container` is memoized while `containerEl` could be passed dynamically via params?

dshovchko: Memoization of $container (and as a three another getters) exist just only during popups update loop. 
It creates after show and clears on hide.
Nothing can open popup again if it is in the open state (if I understand right the design of toggleable).

NastaLeo (approved):

ala-n (approved):

PR #820, opened 2022-02-07
fix(esl-panel): Animation end handling

fshovchko:
In the scope of this PR I fixed issue where function that executes after animation end wasn't firing. The problem was that transition didn't execute when transition property value didn't change. As a result fallback-duration attribute which handled the issue before is now optional and doesn't have default value

### DIFF
--- a/src/modules/esl-panel-group/README.md
+++ b/src/modules/esl-panel-group/README.md
@@ -21,7 +21,7 @@ ESLPanelGroup.register();
 - `mode-cls` - rendering mode class pattern (default: `esl-{mode}-view`). Uses ESLUtils `format` syntax for `mode` placeholder
 - `mode-cls-target` - Element [ESLTraversingQuery](../esl-traversing-query/README.md)  selector to add class that identifies mode (ESLPanelGroup itself by default)
 - `animation-class` - class(es) to be added during animation ('animate' by default)
-- `fallback-duration` - time to clear animation common params (max-height style + classes) (2s by default)
+- `fallback-duration` - time to clear animation common params (max-height style + classes)
 - `no-collapse` - list of comma-separated "modes" to disable collapse/expand animation (for both Group and Panel animations)
 - `accordion-group` - defines accordion behavior: 
   * `single` (default) allows only one Panel to be open
--- a/src/modules/esl-panel-group/core/esl-panel-group.ts
+++ b/src/modules/esl-panel-group/core/esl-panel-group.ts
@@ -31,8 +31,8 @@ export class ESLPanelGroup extends ESLBaseElement {
   @attr({defaultValue: ''}) public modeClsTarget: string;
   /** Class(es) to be added during animation ('animate' by default) */
   @attr({defaultValue: 'animate'}) public animationClass: string;
-  /** Time to clear animation common params (max-height style + classes) ('auto' by default) */
-  @attr({defaultValue: 'auto'}) public fallbackDuration: number | 'auto';
+  /** Time to clear animation common params (max-height style + classes) */
+  @attr() public fallbackDuration: number | 'auto' | '';
   /** List of comma-separated "modes" to disable collapse/expand animation (for both Group and Panel animations) */
   @attr() public noCollapse: string;
   /**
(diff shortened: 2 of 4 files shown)


src/modules/esl-panel-group/README.md
```diff
@@ -21,7 +21,7 @@ ESLPanelGroup.register();
 - `mode-cls` - rendering mode class pattern (default: `esl-{mode}-view`). Uses ESLUtils `format` syntax for `mode` placeholder
 - `mode-cls-target` - Element [ESLTraversingQuery](../esl-traversing-query/README.md)  selector to add class that identifies mode (ESLPanelGroup itself by default)
 - `animation-class` - class(es) to be added during animation ('animate' by default)
-- `fallback-duration` - time to clear animation common params (max-height style + classes) (2s by default)
+- `fallback-duration` - time to clear animation common params (max-height style + classes)
```
ala-n, line 24: The default mentioned originally was not correct, btw.

src/modules/esl-panel-group/core/esl-panel-group.ts
```diff
@@ -31,8 +31,8 @@ export class ESLPanelGroup extends ESLBaseElement {
   @attr({defaultValue: ''}) public modeClsTarget: string;
   /** Class(es) to be added during animation ('animate' by default) */
   @attr({defaultValue: 'animate'}) public animationClass: string;
-  /** Time to clear animation common params (max-height style + classes) ('auto' by default) */
-  @attr({defaultValue: 'auto'}) public fallbackDuration: number | 'auto';
+  /** Time to clear animation common params (max-height style + classes) */
```
ala-n, line 34: Here is the correct definition.
The value 'auto' disables fallback, and it is a default fo component.

src/modules/esl-panel-group/core/esl-panel-group.ts
```diff
@@ -31,8 +31,8 @@ export class ESLPanelGroup extends ESLBaseElement {
...
+  @attr() public fallbackDuration: number | 'auto' | '';
```
ala-n, line 35: No reason for that change see the note above

src/modules/esl-panel-group/core/esl-panel-group.ts
```diff
@@ -233,6 +234,7 @@ export class ESLPanelGroup extends ESLBaseElement {
 
   /** Inits a fallback timer to call post-animate action */
   protected fallbackAnimate(): void {
+    if (!this.fallbackDuration) return;
```
ala-n, line 237: this was already covered by the 'auto' value and 239th line

src/modules/esl-panel/core/esl-panel.ts
```diff
@@ -32,8 +32,8 @@ export class ESLPanel extends ESLToggleable {
   @attr({defaultValue: 'animate'}) public animateClass: string;
   /** Class(es) to be added during animation after next render ('post-animate' by default) */
   @attr({defaultValue: 'post-animate'}) public postAnimateClass: string;
-  /** Time to clear animation common params (max-height style + classes) (1s by default) */
-  @attr({defaultValue: '1000'}) public fallbackDuration: number | 'auto';
+  /** Time to clear animation common params (max-height style + classes) */
+  @attr() public fallbackDuration: number | 'auto' | '';
```
ala-n, line 36: Use `auto` value as a default

src/modules/esl-panel/core/esl-panel.ts
```diff
@@ -104,6 +104,7 @@ export class ESLPanel extends ESLToggleable {
   /** Process animation */
   protected onAnimate(action: string): void {
     // set initial height
+    if (this.initialHeight === 0) this.afterAnimate();
```
ala-n, line 107: Incorrect, just the next line makes me sure that 0 is not the only case here.

src/modules/esl-panel/core/esl-panel.ts
```diff
@@ -128,6 +129,7 @@ export class ESLPanel extends ESLToggleable {
 
   /** Init a fallback timer to call post-animate action */
   protected fallbackAnimate(): void {
+    if (!this.fallbackDuration) return;
```
ala-n, line 132: Same note as for panel-group

ala-n (changes requested):

ala-n:
Covered by #848

PR #821, opened 2022-02-07
feat(esl-utils): Transition duration utility

fshovchko:
In the scope of this PR I added a utility to get a transition length


ala-n (changes requested):
To be discussed on tomorrow's call.

We should have a utility to resolve a class of problems but not specific ones. (But you are really close to making this code universal btw)
Also, the esl-utils can be extended only with unit tests provided for extending, until it's not agreed with a @exadel-inc/esl-core-team.

ala-n:
> To be discussed on tomorrow's call.
> 
> We should have a utility to resolve a class of problems but not specific ones. (But you are really close to making this code universal btw) Also, the esl-utils can be extended only with unit tests provided for extending, until it's not agreed with a @exadel-inc/esl-core-team.

````
/** @example parseTime('.5s') // 500 */
function parseTime(timeStr: string): number;

/** @example parseTimeSet('.5s, 1ms') // [500, 1] */
function parseTimeSet(timeStr: string): number[];
```

src/modules/esl-utils/misc/format.ts
```diff
@@ -67,3 +67,28 @@ export function format(str: string, source: Record<string, any>, matcher: RegExp
...
+/** Parses time, passed in seconds or milliseconds
+ * @example
+ * `.3s`, `4.5s`, `1000ms`
+ * @returns milliseconds
+*/
+export function parseTime(timeStr: string): number {
+  const isMilliseconds = timeStr.indexOf('ms') !== -1;
+  if (isMilliseconds) {
+    timeStr = timeStr.replace('ms', '');
+    return parseNumber(timeStr, NaN)!;
```
ala-n, line 80: Just `+timeStr` is enough in our situation.
Also, it looks like we need to fix a little bit types for `parseNumber` method

ala-n: There is also a coulee minor points (regarding `parseTime('2wtf') // NaN but not 2000`)

ala-n (approved):

src/modules/esl-utils/misc/test/format.test.ts
```diff
@@ -112,4 +114,35 @@ describe('misc/format helper tests', () => {
...
     );
   });
+
+  describe('parseTime', () => {
+    test.each([
+      [['.3s'], 300],
+      [['.124s'], 124],
+      [['4s'], 4000],
+      [['350ms'], 350],
+      [['1024ms'], 1024],
```
dshovchko, line 124: I suggest adding test cases for these values:
4.s
350.ms
Look like these is invalid values.

dshovchko (approved):

NastaLeo (approved):

src/modules/esl-utils/misc/format.ts
```diff
@@ -67,3 +67,23 @@ export function format(str: string, source: Record<string, any>, matcher: RegExp
...
   });
 }
+
+/** Parses time, passed in seconds or milliseconds
+ * @example
+ * `.3s`, `4.5s`, `1000ms`
+ * @returns milliseconds
+*/
+export function parseTime(timeStr: string): number {
+  const isMilliseconds = timeStr.indexOf('ms') !== -1;
```
NastaLeo, line 77: const value = parseFloat(timeStr);
  const isMilliseconds = timeStr.trim().endsWith('ms');
  return value * (isMilliseconds ? 1 : 1000);

dshovchko (approved):

NastaLeo (approved):

julia-murashko (approved):

abarmina (approved):

PR #850, opened 2022-02-19
chore(gh-pages): get rid of _data dir, site data moved to yml

ala-n:
That just an optional proposal just in case )
Feel free to share your opinion

In bounds of that PR:
- move site data to the root for
- convert site.json to site.yml
- move everything under the _data to the 11ty controllers


pages/site.yml
```diff
@@ -0,0 +1,18 @@
+brand: Exadel Smart Library
+description: ESL — A Web Components Library
+
+url: "https://esl-ui.com"
+urlPrefix: ""
```
abarmina, line 5: do you want to keep it instead of deleting it because it can be useful in the future?

abarmina (approved):

NastaLeo (approved):

pages/site.yml
ala-n: Yep, just left it for potential needs in the future. Basically, for the markdown processor (links resolution part), this might appear to be really reasonable.

fshovchko (approved):

PR #908, opened 2022-04-04
fix: event utils passive subscriptions and descriptors merging

ala-n:


NastaLeo (approved):

yadamskaya (approved):

src/modules/esl-utils/dom/events/misc.ts
```diff
@@ -13,6 +13,10 @@ export const getCompositeTarget = (e: CustomEvent): EventTarget | null => {
   return targets ? targets[0] : e.target;
 };
 
+/** See https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md for details */
+const PASSIVE_EVENTS = ['wheel', 'mousewheel', 'touchstart', 'touchmove'];
+export const isPreferPassive = (event: string): boolean => PASSIVE_EVENTS.includes(event);
```
abarmina, line 18: just as a suggestion to name `isPassiveEvent` or something like that

src/modules/esl-utils/dom/events/listener.ts
```diff
@@ -16,8 +17,10 @@ export interface ESLListenerEventMap extends HTMLElementEventMap {
 export type ESLListenerDescriptor<EType extends keyof ESLListenerEventMap = string> = {
   /** Event type (name) */
   event: EType;
-  /** Use capture DOM Event phase */
+  /** Use capture DOM Event phase, see {@link AddEventListenerOptions.capture} for details */
```
abarmina, line 20: cosmetic: can we just use nouns here and below?
```suggestion
  /** Capture DOM Event phase, see {@link AddEventListenerOptions.capture} for details */
```

or mb `using` instead of `use`

abarmina (approved):

src/modules/esl-utils/dom/events/listener.ts
ala-n: Updated using even better version from MDN :) Thx!

src/modules/esl-utils/dom/events/misc.ts
ala-n: Renamed to `isPassiveByDefault` to describe method result more accurate

julia-murashko (approved):

Issue #907, opened 2022-04-04
[🚀esl-media-query]: ability to track breakpoints transformations


ala-n:
As an ESL Media Query consumer I want to be able to 
 - track my project breakpoint changes (e. g. I have 3 main project brekapoints : Small, Medium, Large ; all of them added to the `ESLScreenBreakpoints` registry; I want to be able to track the environment transformations between breakpoints, with an assumption that breakpoints are not intersects)
    ```
     // Detemin the way to subscribe the following method
     onBreakpointChange() {
     }
    ``` 
 - quickly determinate currently active breakpoint (with the same assumption that project breakpoints are not intersects)
    ```
     const currentBp = ESL....; // ?
    ``` 
    Note: currently covered with technically equal checks like `ESLMediaQuery.for('@Small').matches`
    Note: `ESLScreenBreakpoints` can contan much more registrated breakpoints then just the main ones (i.e. it is a wrong assumption to consider ALL `ESLScreenBreakpoints` to be non intersecting)
 - (optional, NTH) be able to receive transformation details, e.g.:
    ```ts
     /**
      * @param current - currently active breakpoint
      * @param previous - breakpoint name that was active before
      */
     onBreakpointChange(current: string, previous: string) {
     }
    ``` 
    Note: the third point looks like the most tricky one in terms of alternatives, please be attentive

## Story output
As the story result is expected to have the following points resolved:
1. Resolution recommendation for cases mentioned in the first part of the story (should be added to the `esl-media-query` documentation) 
2. (If applicable) Small corrections and extensions for `esl-media-query` module
3. (If applicable) Additional API to resolve mentioned problems

As you may notice, the output of the story technically is undefined right now, that means that all the cases may covered by existing functionality (then only samples in "how to" section is needed), or it can be resolved with a list of the small additions to the currentmodule API, or it may require some big API changes. 

Please describe alternatives you've considered with the @exadel-inc/esl-core-team to be sure we chose the right way.

**Priority**: Normal

ala-n:
In bounds of ESL 4.0.0 esl media query capabilities, it is recommended to use shared instance of the `ESLMediaRuleList` to observe breakpoints transformation:

```
// .../shared/breakpoints.ts
export const BREAKPOINTS = new ESLMediaRuleList('@xs|@sm|@md|@lg|@xl', 'xs|sm|md|lg|xl');

// ../some/project-file.ts
class MyComponent {
   @listen({ event: change, target: BREAKPOINTS })
[...]

PR #941, opened 2022-04-23
feat: esl-media-query 3.0 - extended EventTarget based API

ala-n:
`ESLMediaQuery` updated with an `EventTarget` interface support wich allows to use DOM Event Listeners like `ESLEventListener` to observe query changes.

`addLiatener` and `removeListener` methods are deprecated but `addEventLiatener` and `removeEventListener` besides EventTarget API allows to pass single parameter with a callback.

**Open question**: Do we need to support legacy  `addLiatener` and `removeListener` methods with the 4th ESL Major update?
@exadel-inc/esl-all-contributors and consumers what do you think ?

NOTE: PR still in progress: unit tests will be updated, `MediaRuleList` may also get an extended API
#907 is under discussion still, but most likely it's going to have updates related to the new `ESLMediaQuery`  API


src/modules/esl-media-query/core/conditions/media-query-condition.ts
```diff
@@ -48,4 +60,9 @@ export class MediaQueryCondition implements IMediaQueryCondition {
     const complex = inverted && /\)[\s\w]+\(/.test(query);
     return (inverted ? 'not ' : '') + (complex ? `(${query})` : query);
   }
+
+  /** Handle query change and dispatch it on top level in case result value is changed */
```
julia-murashko, line 64:
```suggestion
  /** Handles query change and dispatch it on top level in case result value is changed */
```

src/modules/esl-media-query/core/conditions/media-query-const.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public optimize(): IMediaQueryCondition {
+    return this;
+  }
+  public toString(): string {
+    return this._matches ? 'all' : 'not all';
+  }
+
+  /** Compare const media condition with the passed query instance or string */
```
julia-murashko, line 34:
```suggestion
  /** Compares const media condition with the passed query instance or string */
```

src/modules/esl-media-query/core/conditions/media-query-containers.ts
```diff
@@ -9,40 +8,40 @@ import type {IMediaQueryCondition} from './media-query-base';
...
   public get matches(): boolean {
     return false;
   }
 
-  /** Exclude const conditions. Unwrap empty or trivial (with one item) containers */
-  public optimize(): IMediaQueryCondition {
-    return this;
-  }
-
   /** Handle query change and dispatch it on top level in case result value is changed */
```
julia-murashko, line 40:
```suggestion
  /** Handles query change and dispatch it on top level in case result value is changed */
```

julia-murashko (approved):

src/modules/esl-media-query/core/conditions/media-query-base.ts
```diff
@@ -1,46 +1,82 @@
...
- * Have only two instances: {@link ALL} and {@link NOT_ALL}
- */
-class MediaQueryConstCondition implements IMediaQueryCondition {
-  constructor(private readonly _matches: boolean) {}
-
-  public get matches(): boolean {
-    return this._matches;
+/** Custom event dispatched by {@link ESLMediaQuery} instances */
+export class ESLMediaChangeEvent extends Event {
+  /** `true` if the query is matches device conditions when event was dispatched */
```
dshovchko, line 25:
```suggestion
  /** `true` if the query is matched device conditions when event was dispatched */
```

dshovchko (approved):

src/modules/esl-media-query/core/conditions/media-query-containers.ts
```diff
@@ -9,40 +8,40 @@ import type {IMediaQueryCondition} from './media-query-base';
...
   }
 
-  /** Exclude const conditions. Unwrap empty or trivial (with one item) containers */
-  public optimize(): IMediaQueryCondition {
-    return this;
-  }
-
-  /** Handle query change and dispatch it on top level in case result value is changed */
-  protected _onChildChange(): void {
+  /** Handles query change and dispatch it on top level in case result value is changed */
```
NastaLeo, line 40:
```suggestion
  /** Handles query change and dispatches it on top level in case result value is changed */
```

src/modules/esl-media-query/core/conditions/media-query-condition.ts
```diff
@@ -48,4 +60,9 @@ export class MediaQueryCondition implements IMediaQueryCondition {
     const complex = inverted && /\)[\s\w]+\(/.test(query);
     return (inverted ? 'not ' : '') + (complex ? `(${query})` : query);
   }
+
+  /** Handles query change and dispatch it on top level in case result value is changed */
```
NastaLeo, line 64:
```suggestion
  /** Handles query change and dispatches it on top level in case result value is changed */
```

NastaLeo (approved):

nattallius (approved):

src/modules/esl-media-query/core/conditions/media-query-base.ts
```diff
@@ -1,46 +1,82 @@
...
+export class ESLMediaChangeEvent extends Event {
+  /** `true` if the query is matched device conditions when event was dispatched */
+  public readonly matches: boolean;
+  public readonly target: IMediaQueryCondition;
+
+  constructor(matches: boolean) {
+    if (!window.Reflect) {
+      // ES5 Target
+      super('change');
+      this.matches = matches;
```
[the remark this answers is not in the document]
ala-n: Nope, `return this;` which is equal `return undefined;` in context of operator `new` normal flow can be omitted.
The expression below for browsers that consumes ES6 here is a little bit more complex: it is require fully manual flow to create instance without super syntax direct call, so manually created instance we should pass as the result of consucror function call.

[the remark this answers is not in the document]
ala-n: Thanks @nattallius , now I see  the problem with ES5 target browsers
Will be fixed soon

dshovchko (approved):

NastaLeo (approved):

fshovchko (approved):

Issue #928, opened 2022-04-14
[🚀esl-togglable]: add ability to pass `esl:show:request` additional data


ala-n:
As a ESL consumer, I want to be able to pass additional options with the `esl:show:request`, especially delay to show and filter to ignore some targets.

## API draft
```ts
  interface ESLShowRequestDetails {
      // selector to ignore or exact predicate to check if the target should process request
      ignore: selector | (target: EventTarget) => boolean;
      // Delay to show targets
      delay: number;
      // Custom params to pass
      params: Record<string, any>
  }
```

## Example
```ts
EventUtils.dispatch('esl:show:request', {details: {
   delay: 200,
   ignore: '.speciall-container'
}});
```

Note: please make sure that the original event is also passed to params, see `ToggleableActionParams`

**Priority**: Normal

ala-n:
CC: @NastaLeo

PR #981, opened 2022-05-11
feat(esl-togglable): show:request additional data

fshovchko:
In the scope of this PR I made ability to pass additional options with the esl:show:request

Closes: #928

### DIFF
--- a/src/modules/esl-toggleable/core/esl-toggleable.ts
+++ b/src/modules/esl-toggleable/core/esl-toggleable.ts
@@ -326,14 +326,11 @@ export class ESLToggleable extends ESLBaseElement {
   @bind
   protected _onShowRequest(e: CustomEvent<ESLShowRequestDetails>): void {
     const detail = e.detail;
+    if (detail?.ignore && isMatches(this, detail.ignore)) return;
     const params = {event: e};
-    if (!detail) {
-      this.show(params);
-      return;
-    }
-    if (typeof detail.delay === 'number') Object.assign(params, {showDelay: detail.delay});
-    Object.assign(params, detail.params || {});
-    if (!isMatches(this, detail.ignore)) this.show(params);
+    if (detail && typeof detail.delay === 'number') Object.assign(params, {showDelay: detail.delay});
+    if (detail && typeof detail.parms === 'object') Object.assign(params, detail.params || {});
+    this.show(params);
   }
 }
 
--- a/src/modules/esl-toggleable/core/esl-toggleable.ts
+++ b/src/modules/esl-toggleable/core/esl-toggleable.ts
@@ -326,7 +326,7 @@ export class ESLToggleable extends ESLBaseElement {
   @bind
   protected _onShowRequest(e: CustomEvent<ESLShowRequestDetails>): void {
     const detail = e.detail;
-    if (detail?.ignore && isMatches(this, detail.ignore)) return;
+    if (isMatches(this, detail?.ignore)) return;
     const params = {event: e};
     if (detail && typeof detail.delay === 'number') Object.assign(params, {showDelay: detail.delay});
     if (detail && typeof detail.params === 'object') Object.assign(params, detail.params || {});
--- a/src/modules/esl-toggleable/core/esl-toggleable.ts
+++ b/src/modules/esl-toggleable/core/esl-toggleable.ts
@@ -11,7 +11,7 @@ import {isMatches} from '../../esl-utils/dom/traversing';
 
 export interface ESLShowRequestDetails {
   // Selector to ignore or exact predicate to check if the target should process request
-  ignore?: string | ((target: EventTarget) => boolean);
+  ignore?: string | ((target: Element) => boolean);
   // Delay to show targets
   delay?: number;
   // Custom params to pass
--- a/src/modules/esl-toggleable/core/esl-toggleable.ts
+++ b/src/modules/esl-toggleable/core/esl-toggleable.ts
@@ -324,13 +324,18 @@ export class ESLToggleable extends ESLBaseElement {
   /** Actions to execute on show request */
   @bind
   protected _onShowRequest(e: CustomEvent<ESLShowRequestDetails>): void {
-    const ignore = e.detail.ignore;
     const target = e.target;
     if (!target) return;
-    if (typeof ignore === 'string' && (target as Element).matches(ignore)) return;
-    if (typeof ignore === 'function' && ignore.call(target)) return;
-    const showParams = Object.assign({}, {...e.detail.params} || {}, {delay: e.detail.delay} || {});
-    this.show(showParams);
+    if (!e.detail) {
+      this.show();
+      return;
+    }
+    const ignore = e.detail.ignore;
+    if (typeof ignore === 'string' && (target as Element).matches(ignore) || typeof ignore === 'function' && ignore.call(target)) return;
+    this.show({
+      delay: e.detail.delay || 0,
+      ...e.detail.params
+    });
   }
 }
 
--- a/src/modules/esl-toggleable/core/esl-toggleable.ts
+++ b/src/modules/esl-toggleable/core/esl-toggleable.ts
@@ -7,8 +7,9 @@ import {sequentialUID} from '../../esl-utils/misc/uid';
 import {DeviceDetector} from '../../esl-utils/environment/device-detector';
 import {DelayedTask} from '../../esl-utils/async/delayed-task';
 import {ESLBaseElement, attr, jsonAttr, boolAttr} from '../../esl-base-element/core';
+import {isMatches} from '../../esl-utils/dom/traversing';
 
-interface ESLShowRequestDetails {
+export interface ESLShowRequestDetails {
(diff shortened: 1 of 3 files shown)


src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -8,6 +8,15 @@ import {DeviceDetector} from '../../esl-utils/environment/device-detector';
 import {DelayedTask} from '../../esl-utils/async/delayed-task';
 import {ESLBaseElement, attr, jsonAttr, boolAttr} from '../../esl-base-element/core';
 
+interface ESLShowRequestDetails {
+  // Selector to ignore or exact predicate to check if the target should process request
+  ignore: string | ((target: EventTarget) => boolean);
+  // Delay to show targets
+  delay: number;
```
dshovchko, line 15: It seems to me that ignore and delay are optional params, aren't they?

src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -8,6 +8,15 @@ import {DeviceDetector} from '../../esl-utils/environment/device-detector';
 import {DelayedTask} from '../../esl-utils/async/delayed-task';
 import {ESLBaseElement, attr, jsonAttr, boolAttr} from '../../esl-base-element/core';
 
+interface ESLShowRequestDetails {
```
ala-n, line 11:
```suggestion
export interface ESLShowRequestDetails {
```
We can't predict if that's type will be useful from consumers side, so better let them asses it.

src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -314,8 +323,17 @@ export class ESLToggleable extends ESLBaseElement {
...
   @bind
-  protected _onShowRequest(): void {
-    this.show();
+  protected _onShowRequest(e: CustomEvent<ESLShowRequestDetails>): void {
+    const target = e.target;
+    const detail = e.detail;
+    if (!target) return;
+    let showOptions;
+    if (detail) {
+      showOptions = Object.assign({delay: detail.delay} || {}, {...detail.params} || {});
```
ala-n, line 332: It's better to send exact showDelay in that case.

ala-n: + Why do we need to create a copy here `{...detail.params}` why can't we use it as it is.
`Object.assign` doesn't modify anything except first (target) object

src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -314,8 +323,17 @@ export class ESLToggleable extends ESLBaseElement {
...
+      const ignore = detail.ignore;
+      if (typeof ignore === 'string' && (target as Element).matches(ignore) || typeof ignore === 'function' && ignore.call(target)) return;
```
ala-n, line 334: ignore call doesn't match with signature
we should pass `this` element not an e.target to ignore predicate which is considered as contextless and expects element to be passed as first argument.

+ I'd recommend to simplify this thing, or even move it to utils, something like  
 `isMathces(el: Element, matcher: undefined | selector | (el: Element) => boolean): boolean;`

src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -314,8 +323,17 @@ export class ESLToggleable extends ESLBaseElement {
...
+    }
+    this.show(showOptions);
```
ala-n, line 336: There is nothing bad to send an empty object to `show` just in case

src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -314,8 +323,17 @@ export class ESLToggleable extends ESLBaseElement {
 
   /** Actions to execute on show request */
   @bind
-  protected _onShowRequest(): void {
-    this.show();
+  protected _onShowRequest(e: CustomEvent<ESLShowRequestDetails>): void {
+    const target = e.target;
+    const detail = e.detail;
+    if (!target) return;
+    let showOptions;
```
ala-n, line 330: Oh do we really need a mutable style here?
`let` usually indicates that the code will be overcomlicated instead of splited on the named immutable parts.

ala-n (changes requested):

src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -314,8 +324,12 @@ export class ESLToggleable extends ESLBaseElement {
 
   /** Actions to execute on show request */
   @bind
-  protected _onShowRequest(): void {
-    this.show();
+  protected _onShowRequest(e: CustomEvent<ESLShowRequestDetails>): void {
+    const detail = e.detail;
+    this.show(!isMatches(this, detail.ignore) ?
```
ala-n, line 329: Please recheck the following AC of the story:
 - If the current element doesn't matches passed selector or predicate of `ignore` key then element *skips* `show` processing
 - it is recommended to pass `showDelay` with the `delay` key value if it is numeric
 - `parmas` key can override show params if passed (with an upper priority)
 - if none of `params` and `delay` passed then `show` requesed with an empty `{} or undefined` params

src/modules/esl-utils/dom/traversing.ts
```diff
@@ -13,6 +13,11 @@ export const createSequenceFinder = (next: (el: Element) => Element | null) => {
   };
 };
 
+/** Checks if element matches passed selector or exact predicate function */
+export const isMatches = (el: Element, matcher?: string | ((el: Element) => boolean)): boolean => {
+  return (typeof matcher === 'string' && el.matches(matcher) || typeof matcher === 'function' && matcher.call(el));
+};
```
ala-n, line 19: Matcher predicate still not called according it's signature.
Consider Unit tests to check `isMatches` utility proper work.

ala-n (changes requested):

src/modules/esl-utils/dom/traversing.ts
ala-n: Still not :)
```
const sample = document.createElement('div');
expect(isMatches(sample, (el) => sample === el)).toBe(true);
expect(isMatches(sample, 'div').toBe(true);
expect(isMatches(sample, 'span').toBe(true);
```

src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -314,8 +324,16 @@ export class ESLToggleable extends ESLBaseElement {
...
+  protected _onShowRequest(e: CustomEvent<ESLShowRequestDetails>): void {
+    const detail = e.detail;
+    const params = {event: e};
+    if (!detail) {
+      this.show(params);
+      return;
+    }
+    if (typeof detail.delay === 'number') Object.assign(params, {showDelay: detail.delay});
+    Object.assign(params, detail.params || {});
+    if (!isMatches(this, detail.ignore)) this.show(params);
```
ala-n, line 336:
```suggestion
    const detail = e.detail;
    if (detail?.ignore && isMatches(this, detail.ignore)) return;
    const params = {event: e};
    if (detail && typeof detail.delay === 'number') Object.assign(params, {showDelay: detail.delay});
    if (detail && typeof detail.parms === 'object') Object.assign(params, detail.params || {});
    this.show(params);
```

src/modules/esl-utils/dom/traversing.ts
```diff
@@ -13,6 +13,11 @@ export const createSequenceFinder = (next: (el: Element) => Element | null) => {
   };
 };
 
+/** Checks if element matches passed selector or exact predicate function */
+export const isMatches = (el: Element, matcher?: string | ((el: Element) => boolean)): boolean => {
+  return (typeof matcher === 'string' && el.matches(matcher) || typeof matcher === 'function' && matcher.call(this, el));
```
ala-n, line 18: Id recomend to simply call `matcher(el)` or pass context as null, as `this` reference here is not obvious (and `undefined`, by the way)

Also consider to write checkby cases, it will simplify usage, readability and support. Short version in our case will be easely created by any of popular minifiers.

```suggestion
  if (typeof matcher === 'string') return el.matches(matcher);
  if (typeof matcher === 'function') return matcher(el);
  return false;
```
By the way I missed that case with undefined matcher, then the condition in esl-togglable can be simplified
`if (isMatches(this, detail?.ignore)) return;`
And that's, basically, good note for description of utility that i's falcy for emty matcher

PS: just in case you wander why it's good in terms of support do not combine condition cases here I'd have to say two scenarios:
1) readability: as I mentioned above currently it was not obvious that empty matcher is falcy condition, while in suggested alternative it's pretty clear, it may simplify situation for user with read sources if something not mentioned in docs
2) it's very likely that the `isMatchs` may requested in future for array of matches, alternative representation can be extended easily with alow chance of incorrect implementation.

ala-n (approved):

ala-n (approved):

dshovchko (approved):

yadamskaya (approved):

PR #1001, opened 2022-05-25
feat(esl-utils): ScrollIntoView new implementation

fshovchko:
In the scope of this PR i added promisified version for scrollIntoView().

Parameters:
element - the element which should become visible
options (Optional)

For boolean value:
If true, the top of the element will be aligned to the top of the visible area of the scrollable ancestor. Corresponds to scrollIntoViewOptions: {block: "start", inline: "nearest"}. This is the default value.
If false, the bottom of the element will be aligned to the bottom of the visible area of the scrollable ancestor. Corresponds to scrollIntoViewOptions: {block: "end", inline: "nearest"}.

For Object with the following properties:
behavior (Optional) - defines the transition animation. One of auto or smooth. Defaults to auto.
block (Optional) - defines vertical alignment. One of start, center, end, or nearest. Defaults to start.
inline (Optional) - defines horizontal alignment. One of start, center, end, or nearest. Defaults to nearest.
offsetTop (Optional) - defines vertical offset in pixels. Defaults is 0.
offsetLeft (Optional) - defines horizontal offset in pixels. Defaults is 0.

### DIFF
--- a/src/modules/esl-utils/dom/scroll.ts
+++ b/src/modules/esl-utils/dom/scroll.ts
@@ -1,4 +1,4 @@
-import {tryUntil} from '../async/promise';
+import {createDeferred} from '../async/promise';
 import {getNodeName, getParentNode} from './api';
 
 export type ScrollStrategy = 'none' | 'native' | 'pseudo';
@@ -99,6 +99,55 @@ export function isScrollParent(element: Element): boolean {
   return /auto|scroll|overlay|hidden/.test(overflow + overflowY + overflowX);
 }
 
+
+interface ScrollIntoViewOptionsExtended extends ScrollIntoViewOptions {
+  offsetTop?: number;
+  offsetLeft?: number;
+}
+
+interface StepOptions {
+  deferred: any;
+  behavior: string;
+  el: Element;
+  top: number;
+  left: number;
+  startTime: number;
+  startTop: number;
+  startLeft: number;
+}
+
+interface OffsetParams {
+  setting: string;
+  start: number;
+  length: number;
+  frameStart: number;
+  frameLength: number;
+}
+
+interface OverflowParams {
+  scrollStart: number;
+  elementStart: number;
+}
+
+interface OverflowRecalcParams extends OverflowParams {
+  scrollLength: number;
+  frameLength: number;
+}
+
+interface Rectangle {
+  left: number;
+  top: number;
+  width: number;
+  height: number;
+}
+
+interface Rectangle {
+  left: number;
+  top: number;
+  width: number;
+  height: number;
+}
+
 /**
  * This is a promise-based version of scrollIntoView().
  * Method scrolls the element's parent container such that the element on which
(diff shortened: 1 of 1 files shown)


pages/views/examples/scrollIntoView.njk
```diff
@@ -0,0 +1,35 @@
...
+        <ul class="scrollable-test-lvl-1-wrapper">
+          {% for i in range(1, 17) -%}
+            <li scrollname="{{ i }}" style="background: {{ functions.randomColor() }}" class="scrollable-test-lvl-1">
+              <p class="scrollable-test-p">{{ i }}</p>
+            </li>
+          {%- endfor %}
+        </ul>
+      </div>
+      <esl-scrollbar class="scrollbar-blue" target="::prev(.esl-scrollable-content)"></esl-scrollbar>
+      <esl-scrollbar class="scrollbar-blue scrollbar-second" target="::prev(.esl-scrollable-content)" horizontal></esl-scrollbar>
```
ala-n, line 29: The scroll is only one of the cases
The original issue resolution can not be closed until all the primary cases will be covered.

src/modules/esl-footnotes/core/esl-note.ts
```diff
@@ -7,7 +7,7 @@ import {ESLTooltip} from '../../esl-tooltip/core';
 import {promisifyTimeout, repeatSequence} from '../../esl-utils/async/promise';
 import {EventUtils} from '../../esl-utils/dom/events';
 import {ENTER, SPACE} from '../../esl-utils/dom/keys';
-import {scrollIntoView} from '../../esl-utils/dom/scroll';
+import {scrollIntoView} from '../../esl-utils/misc/scrollIntoView';
```
ala-n, line 10: The module is incorrect `esl-utils/dom/scroll` should be used instead (module name does not depends on the fact of code splitting, see `dom/events` as the reference of structural approach)

src/modules/esl-utils/dom/scroll.ts
```diff
@@ -1,4 +1,4 @@
-import {tryUntil} from '../async/promise';
+import {createDeferred} from '../async/promise';
 import {getNodeName, getParentNode} from './api';
```
ala-n, line 2: The current scroll module should be split in the current or separate request

ala-n (changes requested):
There is a couple of critical structural notes here.

But, basically, I'd want to hold the final review here with the architectural rework first. The final API doesn't look great to me at this point. 
So just let me summarize the tech requironments, I need a little bit more time to rethink the API first, so marking PR as `not ready for merge` right now

ala-n (changes requested):
@fshovchko, please see the use case example pushed to the branch
It looks like `scrollIntoView` currently does not pass successfully for the dynamic content scenario when page scroll is used (the testing wasn't successful on the clients' side for the same reason).

ala-n:
Needs presentational updates from #1022

PR #1008, opened 2022-06-02
refactor(esl-core): ESLMixinElement moved to separate module

ala-n:
BREAKING CHANGE: (beta only) ESLMixinElement now accessible under 'modules/esl-mixin-element/core'


fshovchko (approved):

NastaLeo (approved):

src/modules/esl-mixin-element/README.md
```diff
@@ -0,0 +1,80 @@
+# [ESL](../../../) Mixin Element
+
+Version: *1.0.0*
+
+Authors: *Alexey Stsefanovich (ala'n)*
+
+<a name="intro"></a>
+
+Provides the core for creating custom attribute elements (mixins) and ts decorators to simplify components creation.
```
dshovchko, line 9: Perhaps, should we write typescript instead of ts (from an SEO point of view)?

dshovchko (approved):

abarmina (approved):

Issue #1039, opened 2022-06-16
[💡esl-utils]:  make ScrollUtils functional, update API


ala-n:
As an ESL architect, I want to have ScrollUtils rewritten in a functional style to support tree-shaking optimization. There are also some API changes expected.

API updates proposal:
```
function lock(strategy?: ScrollStrategy, initiatior?: any);
function unlock(initiatior?: any);
```

**Expected behavior/code**
Empty initiator activates force flow, `null` activator saves request-check, non-null initiator passes trough request-check logic

Related to:
https://github.com/exadel-inc/esl/blob/5bf22ac00e6e017b7ebc38fd0a96a30236c08d03/src/modules/esl-utils/dom/scroll/utils.ts#L6

Note: Deprecate but no remove the current ScrollUtility class

PR #1051, opened 2022-06-21
feat(esl-utils): JS based ability to block container scroll

fshovchko:
In the scope of this PR added:
- Ability to lock scroll on the element;
- Ability to remove scroll lock on the element;
- Ability to check if the element has a scroll lock on it.
- Includes #1056: Make ScrollUtils functional, rework API

PR #1056, opened 2022-06-24
refactor(esl-utils): make ScrollUtils functional, update API

Natalie-Smirnova:
Closes: #1039

### DIFF
--- a/src/modules/esl-utils/dom/scroll/utils.ts
+++ b/src/modules/esl-utils/dom/scroll/utils.ts
@@ -3,52 +3,67 @@ export type ScrollStrategy = 'none' | 'native' | 'pseudo';
 const $html = document.documentElement;
 const initiatorSet = new Set();
 
-// TODO: functional
 
+/** Check vertical scroll based on content height */
+export function hasVerticalScroll(target = $html): boolean {
+  return target.scrollHeight > target.clientHeight;
+}
+
+/** Check horizontal scroll based on content height */
+export function hasHorizontalScroll(target = $html): boolean {
+  return target.scrollWidth > target.clientWidth;
+}
+
+/**
+ * Disable scroll on the page.
+ * @param strategy - to make scroll visually disabled
+ * */
+function lockScroll(strategy?: ScrollStrategy): void {
+  const hasScroll = ScrollUtils.hasVerticalScroll();
+  if (strategy && strategy !== 'none' && hasScroll) {
+    $html.classList.add(`esl-${strategy}-scroll`);
+  }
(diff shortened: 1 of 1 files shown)


ala-n (approved):

ala-n:
@Natalie-Smirnova thanks for reworking this API
Just a two-note that needs to be discussed in the nearest weekly:
- it looks like we also need a test for such functionality as well. Going to create a ticket
- it also needs to be discussed if we need to synchronize these updates with the update regarding scroll lock (mentioned as a JS one, but basically it is mostly about an ability to lock container scrolls)

NastaLeo (approved):

fshovchko (approved):

src/modules/esl-utils/dom/scroll/utils.ts
```diff
@@ -3,52 +3,60 @@ export type ScrollStrategy = 'none' | 'native' | 'pseudo';
...
-    initiator && initiatorSet.add(initiator);
-    (initiatorSet.size > 0) && ScrollUtils.lock(strategy);
+/**
+ * Disable scroll on the page.
+ * @param strategy - to make scroll visually disabled
+ * @param initiator - object to associate request with
+ *
+ * TODO: currently requests with different strategy is not taken into account
+ * */
+export function lockScroll(strategy?: ScrollStrategy | null, initiator?: any): void {
```
ala-n, line 32: Draft:
```ts
export function lockScroll(
  target: Element = document.documentElement, 
  options?: {
     strategy?: ScrollStrategy | null, // support for page only for now
     initiator?: any // Lock/unlock only wia the same marker
  }
): void {
```

src/modules/esl-utils/dom/scroll/utils.ts
```diff
@@ -3,52 +3,60 @@ export type ScrollStrategy = 'none' | 'native' | 'pseudo';
...
+
+/** @deprecated use functional utils instead */
+export abstract class ScrollUtils {
+  public static readonly hasVerticalScroll = hasVerticalScroll;
+  public static readonly hasHorizontalScroll = hasHorizontalScroll;
+
+  public static readonly lock = lockScroll;
+  public static readonly unlock = unlockScroll;
+  public static readonly requestLock = (initiator: any, strategy?: ScrollStrategy | null) => lockScroll(strategy, initiator);
+  public static readonly requestUnlock = (initiator: any) => unlockScroll(initiator);
```
ala-n, line 61: As a result of weekly discussio: remove, mark as a breaking change for 4.0.0

ala-n:
Included into: #1051 with the core team discussion update

PR #1090, opened 2022-07-05
refactor: add rules for color hex length and color hex case && change…

Natalie-Smirnova:
- add rules for color hex length and color hex case
- change `color-hex-length` and `color-hex-case` in `.less` files


.stylelintrc.yml
```diff
@@ -83,6 +83,10 @@ rules:
   unit-no-unknown: true
   # Disallow absolute paths for imports in bounds of library files
   esl-less/import-root: never
+  # Disallow long notation for hex colors
+  color-hex-length: short
+  # Disallow uppercase for hex colors
+  color-hex-case: lower
```
dshovchko, line 89: I think it would be easier to understand if the rules were sorted alphabetically

dshovchko (commented):
There is a problem:
sshot_2022-07-05-15-57-30

dshovchko (approved):

NastaLeo (approved):

PR #1093, opened 2022-07-06
refactor(esl-utils): rework public api for SyntheticEventTarget

fshovchko:
In the scope of this PR I reworked public API for SyntheticEventTarget so future changes to the class won't be affecting other components

### DIFF
--- a/src/modules/esl-utils/dom/events/target.ts
+++ b/src/modules/esl-utils/dom/events/target.ts
@@ -4,7 +4,19 @@
  * Ignores event type
  */
 export class SyntheticEventTarget implements EventTarget {
-  protected readonly _listeners = new Set<EventListenerOrEventListenerObject>();
+  private readonly _listeners = new Set<EventListenerOrEventListenerObject>();
+
+  // Parameter "type" will be used in future update
+  public hasEventListener(): boolean;
+  public hasEventListener(type: string | number): boolean;
+  public hasEventListener(type: string, minCount: number): boolean;
+  public hasEventListener(type: string | number = 'change', minCount: number = 0): boolean {
+    if (typeof type === 'number') {
+      minCount = type;
+      type = 'change';
+    }
+    return this._listeners.size > minCount;
+  }
 
   public addEventListener(callback: EventListenerOrEventListenerObject): void;
   public addEventListener(type: 'change', callback: EventListenerOrEventListenerObject): void;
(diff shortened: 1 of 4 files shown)


ala-n:
Basically, I need to ask, @exadel-inc/esl-core-team, 
As you see in the current PR we going to share an ability to get a whole list of observers from EventTarget instance,
but the native implementation deprecates any possibility to access listeners outside debug.
And honestly, I partially see it's reasonable to have subscriptions be encapsulated and hidden from the outside world. 
Just an option we can think about the method to access just the whole state of Target instance, something like
```ts
   this.hasListeners('change'); // or similar short name
```

fshovchko:
> Basically, I need to ask, @exadel-inc/esl-core-team, As you see in the current PR we going to share an ability to get a whole list of observers from EventTarget instance, but the native implementation deprecates any possibility to access listeners outside debug. And honestly, I partially see it's reasonable to have subscriptions be encapsulated and hidden from the outside world.
> 
> Just an option we can think about the method to access just the whole state of Target instance, something like
> 
> ```ts
>    this.hasListeners('change'); // or similar short name
> ```
> 
> From the usage point of view as you can see from the occurrence in the current PR it's more than enough to cover needs of `SyntheticEventTarget`
> 
> So, in case you prefer to enclose observers and go with the conservative `has` method instead of `get` mark this comment with the like +1emojy

this.hasListeners won't brake any of the code that is using SyntheticEventTarget api so I agree to change the method

NastaLeo (approved):

ala-n:
> > Basically, I need to ask, @exadel-inc/esl-core-team, As you see in the current PR we going to share an ability to get a whole list of observers from EventTarget instance, but the native implementation deprecates any possibility to access listeners outside debug. And honestly, I partially see it's reasonable to have subscriptions be encapsulated and hidden from the outside world.
> > Just an option we can think about the method to access just the whole state of Target instance, something like
> > ```ts
> >    this.hasListeners('change'); // or similar short name
> > ```

> >   
> > From the usage point of view as you can see from the occurrence in the current PR it's more than enough to cover needs of `SyntheticEventTarget`
> > So, in case you prefer to enclose observers and go with the conservative `has` method instead of `get` mark this comment with the like +1emojy
> 
> this.hasListeners won't brake any of the code that is using SyntheticEventTarget api so I agree to change the method

According to the discussion results
We don't gonna share inner observers, but just let users know if they are presented.
So yea, @fshovchko, let's do this update

src/modules/esl-utils/dom/events/target.ts
```diff
@@ -4,7 +4,12 @@
  * Ignores event type
  */
 export class SyntheticEventTarget implements EventTarget {
-  protected readonly _listeners = new Set<EventListenerOrEventListenerObject>();
+  private readonly _listeners = new Set<EventListenerOrEventListenerObject>();
+
+  // Parameter "type" will be used in future update
+  public getEventListenersLength(type: string): number {
```
ala-n, line 10:
```ts
  public hasEventListener(type: string, minCount: number = 0): boolean;
```

dshovchko (approved):

abarmina (approved):

NastaLeo (approved):

ala-n (approved):

PR #1119, opened 2022-07-24
style(gh-pages): update site buttons style

ala-n:


Natalie-Smirnova (approved):

pages/src/common/buttons.less
```diff
@@ -34,58 +36,72 @@
...
-    color: #212529;
-    text-decoration: none;
-  }
   &:disabled {
     opacity: 0.65;
   }
 
+  color: var(--btn-color);
+  border: 2px solid var(--btn-color);
+  border-radius: 0.75rem;
```
dshovchko, line 45: Do we need consider adding `number-leading-zero` rule to stylelint config?
I have looked through our less files we use both options with leading zero and without leading zero.

pages/src/common/links.less
```diff
@@ -1,10 +1,12 @@
 @import (reference) "./variables.less";
 
-@arrow-color-escaped: escape(@primary-blue);
-@arrow: "data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 471.2 471.2'%3E%3Cpath fill='@{arrow-color-escaped}' d='M456.57 248.48a18.45 18.45 0 0 0 5.43-12.9 17.88 17.88 0 0 0-5.43-12.9L340.08 106.19a18.34 18.34 0 0 0-25.93 25.93l85.26 85.13H27.53a18.33 18.33 0 1 0 0 36.66h371.88l-85.26 85.26a18.34 18.34 0 0 0 25.93 25.93z'/%3E%3C/svg%3E";
+@arrow: "data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 471.2 471.2'%3E%3Cpath d='M456.57 248.48a18.45 18.45 0 0 0 5.43-12.9 17.88 17.88 0 0 0-5.43-12.9L340.08 106.19a18.34 18.34 0 0 0-25.93 25.93l85.26 85.13H27.53a18.33 18.33 0 1 0 0 36.66h371.88l-85.26 85.26a18.34 18.34 0 0 0 25.93 25.93z'/%3E%3C/svg%3E";
```
dshovchko, line 3: It seems to me that `;charset=utf-8` is redundant attribute in that case.

dshovchko: Should we create a LESS helper function, similar to built-in `data-uri` but that allows pass content instead of file path?
Something like that
```
.data-uri-content(@mime-type, @content) {
  @datauri: %("data:%s,%s", @mime-type, escape(@content));
  @result: url(@datauri);
}
```

dshovchko (approved):

fshovchko (approved):

NastaLeo (approved):

abarmina (approved):

pages/src/common/buttons.less
ala-n: Issue created

pages/src/common/links.less
ala-n: I don't think this will be a frequent case for us, but let's discuss on the nearest tech call

Issue #1132, opened 2022-08-02
[🚀 for esl-panel-group]: API updates to have more control over behavior


NastaLeo:
As ESL consumer, I want to be able to control active panel change and additional behavior depending on the breakpoint.

Original cases:
1)  Mode change behavior
     **Current**: after mode change, ESLPanelGroup activates the initially opened panel 
     **Missed use case**: the ability to safe selection on mode change
2) `no-animate` parm works inconsistently, it's expected to have it selected by the breakpoint instead of mode
3) Panel group goes not control hide actions to prevent tabs mode from containing 0 open panels

The list of updates proposed for esl-panel-group API (according to initial design discussion, thanks to @julia-murashko, @NastaLeo, @ala-n ):
1. [x] Change `no-animate` API to accept media query instead of a list of mods (to have more control over panel behavior)
  It is a breaking change but with a minimum practical impact on users.  
2. [x] Create an attribute `allow-empty` (type of esl-media-query) to allow components to have only empty panels under control.
3. [x] Create an attribute `refresh-state ': typeof `'initial' | 'last' ` to control behavior on mode change.
   - 'initial' - activates initially opened panel(s)
   - 'last' - try to preserve currently active panel(s)
4. [x] replace `accordion-group` with separate mode

ala-n:
1. Change `no-animate` API to accept media query instead of a list of mods
2. Create an attribute `allow-empty : boolean` to allow components to have only empty panels under control
3. Create an attribute `refresh-state : 'initial' | 'last' ` to control behavior on mode change:
   - 'initial' - activates initially opened panel(s)
   - 'last' - try to preserve currently active panel(s)

ala-n:
> As ESL consumer, I want to be able to control active panel change behaviour when the mode was changed according to breakpoint.
> 
> **Current behavoir**: after mode change ESLPanelGroup activates initially opened panel
> 
> **Use case**: ability to safe selection on mode change

Original reuqest.

ala-n:
- Reduce the number of mods to 'accordion' and 'tabs'
- Add attribute `min-open-items` (value : `0 | 1 | number | 'all'` )
- Add attribute `max-open-items` (value: `0 | 1 | 'all'', attribute is ignored by tabs)

PR #1156, opened 2022-08-16
feat(esl-panel-group): add `refresh-state` attribute

NastaLeo:
Part of [#1132](https://github.com/exadel-inc/esl/issues/1132)


src/modules/esl-panel-group/core/esl-panel-group.shape.ts
```diff
@@ -21,6 +21,9 @@ export interface ESLPanelGroupTagShape extends ESLBaseElementShape<ESLPanelGroup
   /** Define a list of comma-separated "modes" to disable collapse/expand animation (for both Group and Panel animations)*/
   'no-animate'?: string;
 
+  /** Define active panels behaviour in case of mode changing. Supported values: `last|initial`*/
+  'refresh-state': string;
```
yadamskaya, line 25: Can it be not mandatory attribute with `initial` value by default?

yadamskaya (approved):

src/modules/esl-panel-group/core/esl-panel-group.shape.ts
NastaLeo: Exactly. Thanks

PR #1173, opened 2022-08-25
refactor(esl-panel-group): refactor mode and limits API

NastaLeo:


ala-n (approved):

pages/views/examples/nested-accordion.njk
```diff
@@ -46,7 +46,7 @@ aside:
 
 <section class="row">
   <div class="col-12">
-    <esl-panel-group class="accordion-group mb-4" accordion-group="multiple">
+    <esl-panel-group class="accordion-group mb-4" minOpenItems="3" maxOpenItems="all">
```
yadamskaya, line 49:
```suggestion
    <esl-panel-group class="accordion-group mb-4" min-open-items="3" max-open-items="all">
```

src/modules/esl-panel-group/core/esl-panel-group.ts
```diff
@@ -238,8 +268,19 @@ export class ESLPanelGroup extends ESLBaseElement {
   protected _onBeforeShow(e: CustomEvent): void {
```
yadamskaya, line 268: I just want to clarify:)
We put an underscore in the method name because it is an event handler? (because as I know protected methods don't need it)

yadamskaya (approved):

pages/views/examples/tabs.njk
```diff
@@ -72,8 +77,15 @@ aside:
...
 <section>
-  <h2>Tabs (with accordion transformation)</h2>
-  <p>Example with a simple tab mode, transforms to accordion on XS and SM</p>
+  <h2>Tabs (with single accordion transformation)</h2>
+  <p>Example with a simple tab mode (restore initial state) that transforms to single accordion on XS and SM (last state)</p>
+  {% code 'html', 'code-block' %}
+    <esl-panel-group class="esl-d-tab-group"
+                     mode="accordion | @+MD => tabs"
+                     max-open-items="1"
+                     refresh-strategy="last | @+MD=> initial"
```
julia-murashko, line 86: space

julia-murashko (approved):

src/modules/esl-panel-group/core/esl-panel-group.ts
ala-n: Private notation in JS is a very interesting thing... It's not only a visibility modifier, but more often, it is the answer to the question: "Is it safe for me to call or use this from outside? (if I don't well known with the inner implementation)"

Here are 2 cases:
1) We definitely can't say it is something public, and we can allow to use it without context what's inside
2) Another thing is that we try not to conflict with the user in case he is going to extend the class using inheritance. So just because it's a listener, it could often be the case...

And I totally understand your question it's something discussable, for example, we have `onShow`, `onHide` lifecycle methods in ESLTogglable (without private notation), but at the same time, we can't say it's something that should be called from the outside...

dshovchko (approved):

src/modules/esl-panel-group/core/esl-panel-group.ts
yadamskaya: @ala-n thanks a lot for the detailed answer!

yadamskaya (approved):

src/modules/esl-panel-group/core/esl-panel-group.ts
ala-n: Always welcome, @yadamskaya :star:

PR #1231, opened 2022-09-20
feat(esl-panel-group): add ability to specify the strategy of closing panels in case the max-open-items limit is reached

### LINKED ISSUE #1190: [🚀esl-panel-group]: add ability to specify the strategy of closing panels in case the `max-open-items` limit is reached
As an ESL consumer, I want to have a predictable and full strategy for esl-panel-group in case it handles `max-open-items` (more than 1 panel).

**Test case:**
 - Configure 4 `esl-panels` inside of `esl-panel-group`
 - Set `esl-panel-group` limit `max-open-items = "3"`
 - Make open 2-d, 3-rd and 4-th panel
 - Try to open 1-panel, repeat reopening of invective panel
 
 **Current behavior:**
 Only 1-st and 2-nd panels can be closed.

**Expected behavior:**
 Strategy to close active panels according to some order, ideally predictable.

Considered solutions:
 1. Implement a round counter to choose an active panel to close 
    - needs to be implemented in POC, not sure it doable with one simple counter
 2. Math random
   - the easiest way, 100% doable, but unpredictable
 3. Some other suggestions....

**Priority** (Minor, Post 4.0.0)

NastaLeo:
Closes: #1190

### DIFF
--- a/src/modules/esl-panel-group/core/esl-panel-group.ts
+++ b/src/modules/esl-panel-group/core/esl-panel-group.ts
@@ -278,17 +278,18 @@ export class ESLPanelGroup extends ESLBaseElement {
   /** Process {@link ESLPanel} pre-show event */
   @listen('esl:before:show')
   protected _onBeforeShow(e: CustomEvent): void {
-    const panel = e.target;
+    const panel = e.detail?.params?.target || e.target;
     if (!this.includesPanel(panel)) return;
 
     const max = this.currentMaxItems;
     const params = this.mergeActionParams({event: e});
+    const hideInitiator = e.detail?.params?.event?.type === 'esl:before:hide';
 
     // All currently active except panel that requested to be open
     const $activePanels = this.$activePanels.filter((el) => el !== panel);
 
     // overflow = pretended to be active (current active + requested panel) - limit
-    const overflow = Math.max(0, $activePanels.length + 1 - max);
+    const overflow = Math.max(0, $activePanels.length + (hideInitiator ? 0 : 1) - max);
     // close all extra active panels (not includes requested one)
     $activePanels.slice(0, overflow).forEach((el) => el.hide(params));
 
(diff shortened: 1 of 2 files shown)


src/modules/esl-panel-group/core/esl-panel-group.ts
```diff
@@ -321,7 +322,11 @@ export class ESLPanelGroup extends ESLBaseElement {
     // activePanels = currentActivePanels - 1 (hide) + 1 if the event produced by 'before:show'
     const activeNumber = this.$activePanels.length - 1 + Number(selfHandled);
 
-    if (activeNumber < this.currentMinItems) return e.preventDefault();
+    if (activeNumber < this.currentMinItems) {
+      if (this.currentMinItems === 1 || this.currentMinItems === this.$panels.length) return e.preventDefault();
+      this.$panels.find(($panel: ESLPanel) => !$panel.open && $panel.show(this.mergeActionParams({event: e, target: $panel})));
```
ala-n, line 327: It's not a good idea to misuse the array methods.
It reduces readability a lot while it does not give any real benefits.

ala-n, line 327: There are a lot of operations here that are not super light, let's discuss possible optimizations and possibly unifications of the algorithm. I believe there is a way to make the first condition as a consequence of the algoritm

ala-n (approved):

abarmina (approved):

yadamskaya (approved):

src/modules/esl-panel-group/core/esl-panel-group.ts
```diff
@@ -283,12 +283,13 @@ export class ESLPanelGroup extends ESLBaseElement {
...
     const params = this.mergeActionParams({event: e});
+    const balanceMarker = e.detail?.params?.event?.type !== 'esl:before:hide';
 
     // All currently active except panel that requested to be open
     const $activePanels = this.$activePanels.filter((el) => el !== panel);
 
-    // overflow = pretended to be active (current active + requested panel) - limit
-    const overflow = Math.max(0, $activePanels.length + 1 - max);
+    // overflow = pretended to be active (current active + balanceMarker (1 if nothing hides)) - limit
+    const overflow = Math.max(0, $activePanels.length + Number(balanceMarker) - max);
```
dshovchko, line 292:
```suggestion
    const overflow = Math.max(0, $activePanels.length + (balanceMarker ? 1 : 0) - max);
```
the ternary solution works faster.

Also you can modify line 281 to
`const {target: panel, detail} = e;`
and 286 line to
`const balanceMarker = detail?.params?.event?.type !== 'esl:before:hide';`

dshovchko (approved):

src/modules/esl-panel-group/core/esl-panel-group.ts
ala-n: "the ternary solution works faster" 
Nobody will see benefits from that, even if it's faster, its less than 0.01ms
And even to test that that's correct, you need to spend a lot of effort (a lot of tests is absolutely lie due to V8 optimizations for routines that happen in "synthetics" tests) ;)

So disagree for that change as a strict rule for our code style.

julia-murashko (approved):

PR #1239, opened 2022-09-25
refactor(esl-utils): unification of large utils name (safe deprecation for old names)

ala-n:


src/modules/esl-event-listener/core/api.ts
```diff
@@ -81,3 +81,6 @@ export class EventUtils {
     return listeners;
   }
 }
+
+/** @deprecated alias for {@link ESLEventListener} */
```
NastaLeo, line 85: If I'm not mistaken:
/** @deprecated alias for {@link ESLEventUtils} */
export const EventUtils = ESLEventUtils;

NastaLeo (approved):

src/modules/esl-traversing-query/core/esl-traversing-query.ts
```diff
@@ -132,18 +132,21 @@ export class TraversingQuery {
...
   }
 }
 
+/** @deprecated alias for {@link ESLTraversingQuery} */
+export const TraversingQuery = ESLTraversingQuery;
+
 declare global {
   export interface ESLLibrary {
-    TraversingQuery: typeof TraversingQuery;
+    TraversingQuery: typeof ESLTraversingQuery;
```
dshovchko, line 150: In global scope we still use ESL.TraversingQuery, don't we?

dshovchko (approved):

src/modules/esl-traversing-query/core/esl-traversing-query.ts
ala-n: Nope, `@ExportNs` stores everything differently (with the name defined in decorator params) and always under ESL "namespace", so globally it will be `ESL.Traversing`

dshovchko: That's exactly what I was asking. Thanks!

abarmina (approved):

yadamskaya (approved):

fshovchko (approved):

PR #1263, opened 2022-10-03
feat(esl-event-listener): improve `ESLEventListener` object to have exact handler for simple subscriptions

ala-n:
For purpose of improving debugging, now event subscription does not wrap handler function if there is no need (`once` marker or delegation (`selector`))

### DIFF
--- a/src/modules/esl-event-listener/core/listener.ts
+++ b/src/modules/esl-event-listener/core/listener.ts
@@ -91,11 +91,12 @@ export class ESLEventListener implements ESLListenerDefinition, EventListenerObj
     return false;
   }
 
-  /** Handles caught event (used as callback for low-level subscriptions) */
-  public handleEvent(e: Event): void {
-    if (!this.isDelegatedTarget(e)) return;
-    this.handler.call(this.host, e, this);
-    if (this.once) this.unsubscribe();
+  /** Returns exact bound handler (used as callback for low-level subscriptions) */
+  @memoize()
+  public get handleEvent(): EventListener {
+    const handlerBound = this.handler.bind(this.host);
+    const handlerFull = this.once ? ((e: Event): void => (handlerBound(e), this.unsubscribe())) : handlerBound;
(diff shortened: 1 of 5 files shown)


src/modules/esl-event-listener/core/listener.ts
```diff
@@ -91,11 +91,12 @@ export class ESLEventListener implements ESLListenerDefinition, EventListenerObj
...
   }
 
-  /** Handles caught event (used as callback for low-level subscriptions) */
-  public handleEvent(e: Event): void {
-    if (!this.isDelegatedTarget(e)) return;
-    this.handler.call(this.host, e, this);
-    if (this.once) this.unsubscribe();
+  /** Returns exact bound handler (used as callback for low-level subscriptions) */
+  @memoize()
+  public get handleEvent(): EventListener {
```
yadamskaya, line 96: Mb it's better to name it now like this:
```suggestion
  public get handlerEvent(): EventListener {
```

yadamskaya (approved):

dshovchko (approved):

src/modules/esl-event-listener/core/listener.ts
ala-n: It is not possible because this name is declared in native EventListenerObject interfaces

src/modules/esl-event-listener/core/listener.ts
```diff
@@ -116,6 +117,7 @@ export class ESLEventListener implements ESLListenerDefinition, EventListenerObj
     const {passive, capture} = this;
     this.unsubscribe();
     memoize.clear(this, '$targets');
+    memoize.clear(this, 'handleEvent');
```
abarmina, line 120:
```suggestion
    memoize.clear(this, 'handlerEvent');
```

abarmina (approved):

NastaLeo (approved):

PR #1266, opened 2022-10-03
feat(esl-animate): Animate mixin element

fshovchko:
In the scope of this pull request i made an ability to use ESLAnimateService using mixin elements. Code is very similar to existing esl-animate custom element implementation.
Also I improved styling on esl-animate example page

### DIFF
--- a/src/modules/esl-animate/core/esl-animate-mixin.ts
+++ b/src/modules/esl-animate/core/esl-animate-mixin.ts
@@ -1,0 +1,97 @@
+import {attr, boolAttr, ESLMixinElement} from '../../esl-mixin-element/core';
+import {ExportNs} from '../../esl-utils/environment/export-ns';
+import {ready} from '../../esl-utils/decorators/ready';
+import {memoize} from '../../esl-utils/decorators/memoize';
+import {ESLTraversingQuery} from '../../esl-traversing-query/core';
+import {parseNumber} from '../../esl-utils/misc/format';
+
+import {ESLAnimateService} from './esl-animate-service';
+
+/**
+ * ESLAnimateMixin - custom mixin element for quick {@link ESLAnimateService} attaching
+ *
+ * Have two types of usage:
+ * - trough the target-s definition, then esl-animate (without it's own content) became invisible plugin-component
+ * `<div esl-animate target="::next"></div><div>Content</div>`
+ * - trough the content wrapping
+ * `<div esl-animate>Content</div>`
+ */
+@ExportNs('AnimateMixin')
+export class ESLAnimateMixin extends ESLMixinElement {
+  public static is = 'esl-animate';
+  public static observedAttributes = ['group', 'repeat', 'target'];
+
+  /**
+   * Class(es) to add on viewport intersection
+   * @see ESLAnimateConfig.cls
+   */
+  @attr({defaultValue: 'in'}) public cls: string;
+
+  /**
+   * Enable group animation for targets
+   * @see ESLAnimateConfig.group
+   */
+  @boolAttr() public group: boolean;
+
+  /**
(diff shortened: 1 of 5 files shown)


ala-n (changes requested):
Unfortunately for the mixin implementation, there is a couple of concerns for current implementation:
1) the main objective of mixin implementation is to be simple+light while observed attributes is a little bit not a "light" part of mixin implementation.
2) mixin is only an additional behavior of a component (any component)...
So in opposition to the custom element, we should not use simple attributes like `group`, `target`, etc. We don't even need to imagine examples of potentially incompatible components that are already all toggleable and triggers... So that does not make mixin helpful...

In the current situation, I'd think about JSON attribute usage instead of multiple attributes... Or at least we should make them specific, and we can make it unobserved in our situation. The lightweight solution is in or situation a priority.
PS: `target` and `$target` should not be considered for mixin implementation at all, animate mixin should always pointing the exact host

src/modules/esl-animate/core/esl-animate-mixin.ts
```diff
@@ -0,0 +1,69 @@
...
+  public disconnectedCallback(): void {
+    super.disconnectedCallback();
+    ESLAnimateService.unobserve(this.$host);
+  }
+
+  /** Reinitialize {@link ESLAnimateService} for targets */
+  public reanimate(): void {
+    const {cls, ratio, repeat} = this.mergeDefaultParams();
+    ESLAnimateService.unobserve(this.$host);
+    memoize.clear(this, '$targets');
```
ala-n, line 66:
```suggestion
```

src/modules/esl-animate/core/esl-animate-mixin.ts
```diff
@@ -0,0 +1,69 @@
...
+ * `<div esl-animate>Content</div>`
+ * Supports aditional parameters^
+ * `<div esl-animate={repeat: true, ratio: 0.8, cls: 'in'}>Content</div>`
+ */
+@ExportNs('AnimateMixin')
+export class ESLAnimateMixin extends ESLMixinElement {
+  public static is = 'esl-animate';
+
+  @jsonAttr({name: ESLAnimateMixin.is})
+  public options: AnimateOptions;
```
ala-n, line 41: It seems that AnimateOptions should be partial here as well, as the consumer should not care to define all of them.
So in that case it seems to be the AnimateOptions type itself should not be restrictive and should be optional out of the box

src/modules/esl-animate/core/esl-animate-mixin.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public static defaultConfig: Partial<AnimateOptions> = {cls: 'in'};
```
ala-n, line 43: Lint: static field should be above non-static members

src/modules/esl-animate/core/esl-animate-mixin.ts
```diff
@@ -0,0 +1,69 @@
...
+    super.disconnectedCallback();
+    ESLAnimateService.unobserve(this.$host);
+  }
+
+  /** Reinitialize {@link ESLAnimateService} for targets */
+  public reanimate(): void {
+    const {cls, ratio, repeat} = this.mergeDefaultParams();
+    ESLAnimateService.unobserve(this.$host);
+    memoize.clear(this, '$targets');
+    ESLAnimateService.observe(this.$host, {force: true, cls, ratio: parseNumber(ratio, 0.4), repeat});
```
ala-n, line 67: Cosmetic: it seems we can move this update into the merge (prepare) Params method.

ala-n (approved):

dshovchko (approved):

PR #1288, opened 2022-10-14
fix(esl-tabs): fix styles based on current scrollable type

### LINKED ISSUE #1287: [🐛 in esl-tabs]: scrollable attribute component styles does not respond to attribute query
`esl-tabs` attribute `scrollable` supports `ESLMediaRuleList` syntax to define its state per media condition.
But default component styles do not respond to the actual `scrollable` state.
`esl-tabs[scrollable]:not([scrollable='disabled']) /* current CSS rule does not respect scrollable media query */`

**To Reproduce**
1. Configure ESLTabs with a scrollable attribute defined with media query
2. Define both `disabled` and any other scrollable state for separate media conditions
3. Observe `esl-tabs` behavior on disabled scrollable media condition

**Current behavior**
Component styles do not switch to a `disabled` state.

**Expected behavior**
Component styles should switch to a `disabled` state when it's actual for defined media condition

**Environment**
- Browser: Chrome, FF, Safari (latest)
- ESL v4.0.0

Note: not reproducible for ESL v3.0.0

NastaLeo:
Closes: #1287


ala-n (approved):

ala-n (commented):
One more small issue presented with the `scrollable` attribute
skipped `scrollable` attribute is equal to `disabled` state
while existing but the empty attribute is equal to the `side` state
The problem is that `this.currentScrollableType` is equal to `''` in that case, but it should be `'side'`

So please update 
`src/modules/esl-tab/core/esl-tabs.ts:50` to
```ts
return this.scrollableTypeRules.activeValue || 'side';
```

ala-n (approved):

abarmina (approved):

yadamskaya (approved):

dshovchko (approved):

julia-murashko (approved):

PR #1307, opened 2022-10-24
[gh-pages]: UIPlayground initial integration

ala-n:
Simple UIPlayground with Media


pages/src/localdev.ts
```diff
@@ -104,3 +103,6 @@ ESLCarouselPlugins.Dots.register();
 ESLCarouselPlugins.Link.register();
 ESLCarouselPlugins.Touch.register();
 ESLCarouselPlugins.Autoplay.register();
+
+// UI Playground
+import './ui-playground/ui-playground';
```
ala-n, line 108: Still a part of the main bundle build
We should make sure it will load separately (async external load or load independently from template itself)

Issue #931, opened 2022-04-14
[🚀esl-utils]: (POC) create `isVisible` predicate


ala-n:
As an ESL consumer, I want to be able to check element visibility using ESL utils.

**Suggested implementation**
`elem.offsetHeight || elem.offsetWidth || elem.getClientRects().length`

**Additional notes*
Consider the option to include elements with `visibility: hidden` or, in general, the option to choose visibility check strategy

** Nice To Have (follow-ups)**
- [ ] TraversingQuery: support of `::visible` syntax

ala-n:
Split into 2 story:
- [ ] POC: investigate options to check visibility, create isVisible predicate under esl-utils
- [ ] TaaversinfQuery: support `::visible` syntax

ala-n:
@ala-n to write spec

PR #1341, opened 2022-11-21
feat(esl-utils): isVisible predicate

fshovchko:
In the scope of this PR:
- function isVisible to Check if specified element is visible. By default check similar to jQuery is ran in combination with check for presence of element in viewport
- function findParentByStyles to find element's parent with specified styles
#931


dshovchko (commented):
I see that the method checks if there is element in viewport. But I don't think it's enough. Element can be in the viewport. Its scroll parent can be in the viewport. But element can be out of the view area of the scroll parent.

ala-n (commented):
Looks awesome!

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (getNodeName(el) === 'html') return undefined;
+  return findParentByStyles(getParentNode(el) as HTMLElement, matchStyles);
+}
+
+/**
+ * Check if specified element is visible
+ * @param el - element to be checked
+ * @param options - object of additional visibility options to include
+ */
+export function isVisible(el: HTMLElement, options: VisibilityOptions = {visibility: true, opacity: false, viewport: false, size: false}): boolean {
```
ala-n, line 34:
```suggestion
export function isVisible(el: HTMLElement, options: VisibilityOptions = {visibility: true}): boolean {
```

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Get the parent of the specified element with matching styles
+ * @param el - element for which to get the parent
+ * @param matchStyles - object of styles for comparing
+ */
+export function findParentByStyles(el: HTMLElement, matchStyles: Partial<CSSStyleDeclaration>): HTMLElement | undefined {
+  const styleNames = Object.keys(matchStyles);
+  if (!styleNames.length) return undefined;
+
+  const elStyle = window.getComputedStyle(el);
+  if (styleNames.some((styleName: any) => elStyle[styleName] === matchStyles[styleName])) return el;
```
ala-n, line 24: Not obvious point: we don't  declare that the operator-combinator is  OR
By default, if nothing special is mentioned, we assume that the combination will use AND operator 
e.g. find(el, {a: 1, b: 2}) means find element with property a:1 AND b: 2
We need to think how to deal with that conceptual problem (make it configurable, change check strategy or hide method from outside world)

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,48 @@
...
+  /** Element will be considered invisible if it's height or width is '0' */
+  size?: boolean;
+}
+
+/**
+ * Get the parent of the specified element with matching styles
+ * @param el - element for which to get the parent
+ * @param matchStyles - object of styles for comparing
+ */
+export function findParentByStyles(el: HTMLElement, matchStyles: Partial<CSSStyleDeclaration>): HTMLElement | undefined {
```
ala-n, line 19: Method placement doesn't look fine
If we make it public, we should consider some other place for it (e.g., traversing)
Also we need to cover such things by tests if possible

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,48 @@
...
+  const {visibility, opacity} = options;
+  if (!(visibility || opacity)) return true;
+  const parentStyles = Object.assign({}, visibility && {visibility: 'hidden'}, opacity && {opacity: '0'}) as Partial<CSSStyleDeclaration>;
+  return !findParentByStyles(el, parentStyles);
+}
+
+function isInViewport(el: HTMLElement): boolean {
+  const elrect = el.getBoundingClientRect();
+  return el.isSameNode(document.elementFromPoint(elrect.x, elrect.y));
+}
```
dshovchko, line 48: To discuss: is it correct to check the top left corner of the element? Maybe we should check the center of the element?

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Get the parent of the specified element with matching styles
+ * @param el - element for which to get the parent
+ * @param matchStyles - object of styles for comparing
+ */
+export function findParentByStyles(el: HTMLElement, matchStyles: Partial<CSSStyleDeclaration>): HTMLElement | undefined {
+  const styleNames = Object.keys(matchStyles);
+  if (!styleNames.length) return undefined;
```
dshovchko, line 21: Just return will be enough
```suggestion
  if (!styleNames.length) return;
```

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const elStyle = window.getComputedStyle(el);
+  if (styleNames.some((styleName: any) => elStyle[styleName] === matchStyles[styleName])) return el;
+  if (getNodeName(el) === 'html') return undefined;
```
dshovchko, line 25:
```suggestion
  if (getNodeName(el) === 'html') return;
```

src/modules/esl-utils/dom/visible.ts
```diff
@@ -43,6 +28,20 @@ export function isVisible(el: HTMLElement, options: VisibilityOptions = {visibil
...
+}
+
+function computeRectIntersection(rect1: DOMRect | Rect, rect2: DOMRect | Rect): Rect {
+  const top = Math.max(rect1.top, rect2.top);
+  const left = Math.max(rect1.left, rect2.left);
+  const bottom = Math.min(rect1.bottom, rect2.bottom);
+  const right = Math.min(rect1.right, rect2.right);
+  const width = right - left;
+  const height = bottom - top;
+  return Rect.from({top, left, width, height});
```
dshovchko, line 46: It seems to me that it should be a static member of Rect class.

ala-n (changes requested):

src/modules/esl-utils/dom/rect.ts
```diff
@@ -112,4 +127,11 @@ export class Rect {
   public shrink(decrementX: number, decrementY: number = decrementX): Rect {
     return this.grow(-decrementX, -decrementY);
   }
+
+  /**
+   * Method to check if element's area is bigger than zero
+   */
+  public isEmpty(): boolean {
+    return this.height <= 0 || this.width <= 0;
```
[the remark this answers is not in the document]
fshovchko: Nope, it's OR condition on purpose, because there are cases when only one value is negative or equal to zero. Check test cases for example example
` test.each([
    [[0, 0, 10, 10], false],
    [[0, 0, -10, 10], true],
    [[0, 0, 10, -10], true],
    [[0, 0, 0, 0], true],
    [[0, 0, -10, -10], true]
  ])('isEmpty: %s', (coords, expected) => {
    expect(new Rect(...coords).isEmpty()).toBe(expected);
  });`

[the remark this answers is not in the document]
ala-n: Agree in that case that the emptiness of the rectangle is a really undefined term
is the point an empty rectangle - most likely, yes
is the line an empty rectangle - not sure 100%; depends on terminology or applicable task

To resolve the unclearness, propose moving to the square term.
In our case element is not visible when geometrically: the square of intersecting viewport and element rectangle is 0.

PS: negative values for width and height is not valid values that we will not see in reality, or in another point of view it is a valid inverse rectangle but it will not have a 0 square the same as it is not empty )

ala-n (changes requested):

src/modules/esl-utils/dom/rect.ts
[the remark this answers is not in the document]
fshovchko: The main reason of implementing isEmpty was to check if rectangle's area returned by Rect.intersect() is equal to zero or returned rectangle was sort of 'inverted'. Maybe, we can handle inversion differently, for example by getting max value between 0 and dimension values
```
const width = Math.max(right - x, 0);
const height = Math.max(bottom - y, 0);
```

yadamskaya (approved):

src/modules/esl-utils/dom/rect.ts
```diff
@@ -28,6 +28,21 @@ export class Rect {
...
   }
 
+  /**
+   * The static method returns intersection Rect of two Rect instances
+   * @param rect1 - first instance of Rect
+   * @param rect2 - second instance of Rect
+   */
+  public static intersect(rect1: Rect, rect2: Rect): Rect {
+    const top = Math.max(rect1.top, rect2.top);
+    const left = Math.max(rect1.left, rect2.left);
```
dshovchko, line 38: Probably should use `y` instead of `top` and `x` instead of `left` because they are aliases of `x` and `y` properties.

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param el - element to be checked
+ * @param options - object of additional visibility options to include
+ */
+export function isVisible(el: HTMLElement, options: VisibilityOptions = {visibility: true}): boolean {
+  if (!el.getClientRects().length) return false;
+  if (options.visibility && findHost(el, (host) => getComputedStyle(host).visibility === 'hidden')) return false;
+  if (options.opacity &&
+    findClosestBy(el, (parent) => parent instanceof Element && getComputedStyle(parent).opacity === '0')) return false;
+  return !(options.viewport && !isInViewport(el));
+}
```
dshovchko, line 26: Is it correct that `isVisible(el, {})` always returns `true`?

dshovchko (approved):

dshovchko (commented):
Please update docs
sshot_2023-01-11-23-01-35

src/modules/esl-utils/dom/rect.ts
fshovchko: This was discussed with @ala-n and we decided to keep top/left for consistency with bottom/right

src/modules/esl-utils/dom/visible.ts
fshovchko: Not always. Function will always check for display css property

dshovchko (approved):

abarmina (approved):

src/modules/esl-utils/dom/rect.ts
```diff
@@ -112,4 +127,11 @@ export class Rect {
   public shrink(decrementX: number, decrementY: number = decrementX): Rect {
     return this.grow(-decrementX, -decrementY);
   }
+
+  /**
+   * Method to check if element's area is bigger than zero
+   */
+  public isEmpty(): boolean {
+    return this.height === 0 || this.width === 0;
+  }
```
ala-n, line 136: Recommended proposal:
```suggestion
  /** @returns the rectangle area square */
  public get square(): boolean {
    return this.height * this.width;
  }
```
With the consumption:
```
 rect.square === 0 // instead rect.isEmpty()
```

As an alternative:
```suggestion
  /** @returns if the element's area is bigger than zero */
  public get empty(): boolean {
    return this.height === 0 || this.width === 0;
  }
```

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,33 @@
...
+
+function isInViewport(el: HTMLElement): boolean {
+  const wndIntersection = Rect.intersect(getWindowRect(), Rect.from(el.getBoundingClientRect()));
+  if (wndIntersection.isEmpty()) return false;
+  return !getListScrollParents(el).some((parent: HTMLElement) =>
+    Rect.intersect(wndIntersection, Rect.from(parent.getBoundingClientRect())).isEmpty());
```
ala-n, line 32:
```suggestion
  if (wndIntersection.square === 0) return false;
  return !getListScrollParents(el).some((parent: HTMLElement) =>
    Rect.intersect(wndIntersection, Rect.from(parent.getBoundingClientRect())).square === 0);
```
or
```suggestion
  if (wndIntersection.empty) return false;
  return !getListScrollParents(el).some((parent: HTMLElement) =>
    Rect.intersect(wndIntersection, Rect.from(parent.getBoundingClientRect())).empty);
```

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export function isVisible(el: HTMLElement, options: VisibilityOptions = {visibility: true}): boolean {
+  if (!el.getClientRects().length) return false;
+  if (options.visibility && findHost(el, (host) => getComputedStyle(host).visibility === 'hidden')) return false;
+  if (options.opacity &&
+    findClosestBy(el, (parent) => parent instanceof Element && getComputedStyle(parent).opacity === '0')) return false;
+  return !(options.viewport && !isInViewport(el));
+}
+
+function isInViewport(el: HTMLElement): boolean {
```
ala-n, line 28:
```suggestion
/**
 * Checks if the specified element is inside the viewport
 * @param el - element to be checked
 */
export function isInViewport(el: HTMLElement): boolean {
```

src/modules/esl-utils/dom/visible.ts
```diff
@@ -0,0 +1,33 @@
...
+  /** Element will be considered invisible if opacity set to '0' */
+  opacity?: boolean;
+  /** Element will be considered invisible if visibility set to 'hidden' */
+  visibility?: boolean;
+  /** Element will be considered invisible if it's not in viewport */
+  viewport?: boolean;
+}
+
+/**
+ * Check if specified element is visible
```
ala-n, line 16:
```suggestion
 * Checks if the specified element is visible
```

ala-n (approved):
@exadel-inc/esl-core-team and @exadel-inc/esl-all-contributors 
just a small stylistic API topic to vote on and apply before closing this epic.
In the comments below are two options described to check the Rect area with 0 from user perspective:
 - `if (rect.square === 0)`   - 🚀 
 - `if (rect.empty)`               - 👍🏻
 
Please vote on the final implementation under the current message with emoji.

PS: That's the only questions that are left for the current PR before release. 
All tests are passed super successfully, the feature is finally approved for the 4.1 release 🎉

NastaLeo (approved):

Natalie-Smirnova (approved):

ala-n:
Sorry team @exadel-inc/esl-all-contributors , just in case I fixed my incorrect English term in the voting above, so, please take part and share your thoughts ;)

andreybelous:
> @exadel-inc/esl-core-team and @exadel-inc/esl-all-contributors just a small stylistic API topic to vote on and apply before closing this epic. In the comments below are two options described to check the Rect area with 0 from user perspective:
> 
> * `if (rect.area === 0)`   - 🚀
> * `if (rect.empty)`               - 👍🏻
> 
> Please vote on the final implementation under the current message with emoji.
> 

Hi! I'd vote for the `empty` option, but have a feeling that the `area` property could be helpful for consumers as well.

NastaLeo:
I vote for `isEmpty` implementation with `hasArea` name

ala-n:
The final decision for the question above was finalized (on regular team call): `rect.area` property + `rect.area === 0` check
Both properties was the most relevant suggestion, but as we are developing client-side library, we should keep it small so wins the most universal implementation

ala-n (approved):

dshovchko (approved):

Issue #1354, opened 2022-11-30
[🚀esl-utils]: ability to clean multiple memorized fields using simplified compact call


ala-n:
As an ESL consumer, I'd like to have the ability to clear multiple memorized fields in a simplified manner.

Suggestion draft:

Current:
```
  memoize.clear(this, 'field1');
  memoize.clear(this, 'field2');
  memoize.clear(this, 'field3');
  memoize.clear(this, 'field4');
```

Suggested: 
```
  memoize.clear(this, ['field1', 'field2', 'field3', 'field4']);
```

PR #1360, opened 2022-12-03
feat(esl-utils): simplified memoize clear

fshovchko:
This PR adds ability to clear multiple memorized fields in a simplified manner.
#1354

### DIFF
--- a/src/modules/esl-utils/decorators/memoize.ts
+++ b/src/modules/esl-utils/decorators/memoize.ts
@@ -1,5 +1,5 @@
 import {defaultArgsHashFn, memoizeFn} from '../misc/memoize';
-import {isPrototype, getPropertyDescriptor} from '../misc/object';
+import {isPrototype, getPropertyDescriptor, isArrayLike} from '../misc/object';
 
 import type {MemoHashFn} from '../misc/memoize';
 import type {MethodTypedDecorator} from '../misc/functions';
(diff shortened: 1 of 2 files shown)


src/modules/esl-utils/decorators/memoize.ts
```diff
@@ -62,7 +62,9 @@ function memoizeMethod(originalMethod: any, prop: string, hashFn: MemoHashFn) {
  * Note: be sure that you targeting memoized property or function.
  * Clear utility has no 100% check to prevent modifying incorrect (not memoized) property keys
  */
-memoize.clear = function (target: any, property: string): void {
+memoize.clear = function (target: any, property: string | string[]): void {
+  if (isArrayLike(property)) return (property as string[]).forEach((prop) => memoize.clear(target, prop));
```
ala-n, line 66:
```suggestion
  if (Array.isArray(property)) return property.forEach((prop) => memoize.clear(target, prop));
```
Keep it simple ;)
TS is notifying you that something is potentially not ok when the type was requested to be cast.
First things first, we declared on the 65th line that it will be an array but not an ArrayLike, and that's correct (according to the requirements, that's exactly what was requested)
Then the `isArrayLike` is used. If we check the definition and declaration of that predicate, it appears to be a real type guard that checks if the object is array-like (has length and indexed attributes) (and declares that thing `ArrayLike`). But as we clarified in the previous line, we expect it to be an exact array, but not an array-like object.
Then we were requested to cast the type to an array, but why if we've defined the type correctly? Everything goes as expected: `isArrayLike` is a type guard for an `ArrayLike`, so it is not an array, and it is not required to have `Array.prototype` in the proto-chain, so that means if its real `ArrayLike` it can possibly have no `forEach` method, that what TS was notifying us before we cast type back to an array.

ala-n (changes requested):

ala-n (approved):

NastaLeo (approved):

dshovchko (approved):

abarmina (approved):

Natalie-Smirnova (approved):

Issue #427, opened 2021-08-27
[🚀esl-panel-group]: add ability to collapse inner accordion when nesting accordion


julia-murashko:
As an esl-panel-group component consumer, I want to have the ability to support inner accordion with the following behavoir:
hide event of panel is followed by all its nested panels collapsing.

Implement ESLNestedPanelGroup (ESLNestedAccordion) mixin

ala-n:
Recommended spec (suggested bt @ala-n , @NastaLeo, @Natalie-Smirnova ):

Extend `esl-related-target` mixin.
Scope of updates:
 - [ ] make related target able to target multiple related togglable
 - [ ] make related target mixin customizable in terms of observed actions:
    introduce `esl-related-target-action` with 'hide'|'show'|'all'(default) as a values

PR #1380, opened 2022-12-14
feat(esl-panel-group): add ability to collapse inner accordion when nesting accordion

Natalie-Smirnova:
The ability to collapse inner accordion when nesting accordion is added

Closes: #427

### DIFF
--- a/pages/views/examples/accordion.njk
+++ b/pages/views/examples/accordion.njk
@@ -6,14 +6,14 @@ name: Accordion
 tags: examples
 icon: examples/accordion.svg
 aside:
-  components:
-    - esl-panel
-    - esl-panel-group
-    - esl-trigger
-    - esl-toggleable
-  examples:
-    - tabs
-    - toggleable
+components:
+- esl-panel
+- esl-panel-group
+- esl-trigger
+- esl-toggleable
+examples:
+- tabs
+- toggleable
 ---
 {% import 'lorem.njk' as lorem %}
 
@@ -24,7 +24,7 @@ aside:
       <esl-panel-group class="accordion-group mb-4"/>
     {% endcode %}
 
-    <esl-panel-group class="accordion-group mb-4" >
+    <esl-panel-group class="accordion-group mb-4">
       <div class="esl-d-accordion">
         <esl-trigger class="esl-d-accordion-header h4 esl-d-accordion-header-arrow" target="::next">
           Toggle Accordion #1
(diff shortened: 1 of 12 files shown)


src/modules/esl-nested-accordion/core/esl-nested-accordion.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @author Anastasia Lesun, Natalia Smirnova
+ *
+ * ESLNestedAccordion is a custom mixin element that can be used with {@link ESLPanel} instance to allow hide all children panels
+ */
+@ExportNs('NestedAccordion')
+export class ESLNestedAccordion extends ESLMixinElement {
+  static is = 'esl-nested-accordion';
+
+  /** @returns all nested panels  */
+  public get $nestedPanels(): ESLPanel[] {
```
ala-n, line 17: I don't think collecting all the panels inside is a good idea.
At least we never previously in ESL restrict users to changing the target.
We need to think about `ESLTraversingQuery` usage here instead

src/modules/esl-nested-accordion/core/esl-nested-accordion.ts
```diff
@@ -0,0 +1,27 @@
...
+    return [...this.$host.querySelectorAll(ESLPanel.is)] as ESLPanel[];
+  }
+
+  /** Processes {@link ESLPanel} hide event */
+  @listen('esl:hide')
```
ala-n, line 22: Are we 100% sure that the `esl:hide` event was bubbled from the right target?

src/modules/esl-nested-accordion/test/nested-accordion-template.html
```diff
@@ -0,0 +1,37 @@
+<esl-panel-group max-open-items="all" min-open-items="0">
+<div class="esl-d-accordion">
```
ala-n, line 2: Cosmetic: indents (reformat)

src/modules/esl-nested-accordion/test/esl-nested-accordion.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import {ESLTrigger} from '../../esl-trigger/core/esl-trigger';
+
+describe('ESLNestedAccordion: nested panels hide if parent panel (with esl-nested-accordion) is closing', () => {
+
+  ESLPanelGroup.register();
+  ESLTrigger.register();
+  ESLPanel.register();
+  ESLNestedAccordion.register();
+
+  document.body.innerHTML = fs.readFileSync(path.resolve(__dirname, 'nested-accordion-template.html'), 'utf8');
```
ala-n, line 15: Why do we use `fs.read` instead of `html-loder-jest` here?

pages/views/examples/accordion.njk
```diff
@@ -184,3 +184,118 @@ aside:
     </esl-panel-group>
   </div>
 </section>
+
+<section class="row">
```
ala-n, line 188: TODO: discuss example placement

ala-n (changes requested):

PR #1387, opened 2022-12-27
Fix redirect for `/utils/esl-utils/`

Natalie-Smirnova:
The permalink for `/utils/esl-utils/` was added

### DIFF
--- a/pages/views/redirect/esl-utils.njk
+++ b/pages/views/redirect/esl-utils.njk
@@ -1,6 +1,6 @@
 ---
 layout: forward
-target: /core/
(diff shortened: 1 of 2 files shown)


ala-n (approved):

pages/views/redirect/esl-utils.njk
```diff
@@ -0,0 +1,6 @@
+---
+layout: forward
+target: /core/
```
ala-n, line 3:
```suggestion
target: /core/esl-utils/
```

NastaLeo (approved):

PR #1437, opened 2023-01-25
feat(esl-event-listener): bugfix and implementation rework of ESLEventListener descriptors

### LINKED ISSUE #1428: [🐛❗ in esl-event-listener]: performance gap in esl-event-listener auto subscription
The ESL Event Listener API provides `ESLEventUtils.descriptors` collector to get all the descriptors properties from the object created using the `@listen` annotation. The API is used by `ESLMixinElement` and `ESLBaseElement` as well as it's publically accessible to collect all the descriptors that should provide auto-subscribed ESLEventListener subscriptions.

The current `ESLEventUtils.descriptors` implementation causes a performance issue (reflow and low time of execution) due to `extractValues` utility implementation. 

**Current behavior**
`ESLEventUtils.descriptors` visits all of the enumerable properties that exist on the object, which causes:
 - reflow (due to DOM API getters handling (e.g. `offsetWeight`, `offsetHeight`))
 - causes early initialization of lazy getters as well as accessing heavy getters

**Expected behavior**
`ESLEventUtils.descriptors` should execute fast and do not leas to reflow.

**Additional context**
Basically, there is a 'small' in @listenr implementation as it produces an enumerable property. Potentially that should be handled differently.

**Possible Solution**
Make descriptors auto-collection through the middleware storage (hidden list, or any other similar way) instead of passing through the whole list of object keys and values.

ala-n:
In the scope of the current issue:
 - [x] Reworked descriptors API in `ESLEventUtils`
 - [x] Low level refactoring 
 - [x] Fixed #1428 (performance issue with the auto subscription mechanism)
 - [x] Simplified and moved to the `esl-event-listener` module implementation of `@listen`
 - [x] Tests for descriptors API
 - [x] `esl-event-listener` internal implementation uses symbols keys if it's accessable

**WIP: result API, naming, and inner implementation details are still in progress. PR is opened to provide the visibility of the general concept of update** 

Closes: #1428


ala-n:
@exadel-inc/esl-all-contributors , I think there is a moment to discuss if there is a concern with the API changes happening in the current PR.

There is a short summary of them:
 - `descriptors` no longer support any additional params (auto filter previously (not working)), now there is the only goal of that method - resolve all auto-subscribable descriptors of the host object
 - `descriptors` is renamed to the `getAutoDescriptors` (❔), `descriptors` now marked as deprecated
 - `listen` decorator now have a simple alternative implementation - `ESLEventUtils.initDescriptor` (❔) to make the function key of the object a descriptor (encapsulates logic of auto descriptors store and `inherit` key handling).
 
 - (💡tech ) `listen` decorator is now super simplified (due to moving implementation to the `ESLEventUtils.initDescriptor`)
 - (💡tech ) keys for auto-decorators now stored with a hidden property '__esl_decorators' (symbol for ES6)
 
 WHAT IS NOT IMPORTANT and ready for the current feedback - implementation details (typos, internal function naming, and low-level implementation), tests 
 
MAJOR CHANGE FROM INTERNAL PRE-DISCUSSION: there is no public `initAutoDescriptor` at all due to `initDescriptor` public method and also considering auto-subscribed own property key as not realistic.

Feel free to share your thoughts here in the PR or in internal chats all the points and decisions will be reflected here

ala-n:
And also just a tiny additional topic:
Now we have user cases of the custom initialization of all the class auto-descriptors. 
So what do you think? Should we consider a shortcut for the following operation?
`ESLEventUtils.descriptors(this).forEach((desc) => ESLEventUtils.subscribe(this, desc));`

For example just `ESLEventUtils.subscribe(this);` special behaviour 
Or smth like `ESLEventUtils.aubscribeAuto(this);`

src/modules/esl-event-listener/core/descriptors.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/** Gets {@link ESLListenerDescriptorFn}s of the passed object */
+export function getAutoDescriptors(host: unknown): ESLListenerDescriptorFn[] {
+  if (!isObject(host)) return [];
+  const keys = getDescriptorsKeysFor(host);
+  const values = keys.map((key) => host[key]);
+  return values.filter((desc: ESLListenerDescriptorFn) => isEventDescriptor(desc) && desc.auto === true);
+}
+
+/** Mark field, instanceof {@link ESLListenerDescriptorFn}, as collectable event descriptor */
```
fshovchko, line 33:
```suggestion
/** Marks field, instanceof {@link ESLListenerDescriptorFn}, as collectable event descriptor */
```

fshovchko (approved):

dshovchko (approved):

ala-n:
Update PR with experimental implementation, low-level refactoring, and tests

1. Introduce separate `ESLEventUtils.subscribeAll` (the name is under discussion) method to subscribe all auto-subscribable decorators
2. `setAutoDescriptor` removed at all, `ESLEventUtils.initDescriptor` can be used instead
3. `ESLEventUtils.initDescriptor` allows re-declaration
4. `ESLEventUtils.initDescriptor` failed if no parent descriptor was found for `inherit: true` descriptor meta
5. Restrict `ESLEventUtils.initDescriptor` signature for the description meta-object (move shortcut normalization back to `@listen`)

CC: @exadel-inc/esl-all-contributors

dshovchko (approved):

ala-n:
According to final discussion call the only last change to the current state will be moving `subscribeAll` under overload of the `subscribe` method, e.g. :
```
  ESLEventUtils.sunscribe(host); // subscribes all auto descriptors of the host
```

dshovchko (approved):

ala-n:
Documentation update will go just after that update in the separate PR

NastaLeo (approved):

yadamskaya (approved):

abarmina (approved):

src/modules/esl-event-listener/core/descriptors.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @param host - host object
+ * @param createIfNotExists - create keys store on the host if not exists
+ * @returns object own descriptors keys or an empty array
+ */
+function getOwnDescriptors(host: object, createIfNotExists = false): string[] {
+  if (Object.hasOwnProperty.call(host, DESCRIPTORS)) return (host as any)[DESCRIPTORS];
+  const value: string[]  = [];
+  if (createIfNotExists) Object.defineProperty(host, DESCRIPTORS, {value, configurable: true});
```
andreybelous, line 16: cosmetic: it would probably be a bit more transparent to define the property in the separate function

src/modules/esl-event-listener/test/descriptor.test.ts
```diff
@@ -0,0 +1,177 @@
+import {ESLEventUtils} from '../core';
+
+describe('dom/events: ESLEventUtils: ESLListenerDescriptor Utils', () => {
+  describe('ESLEventUtils.isEventDescriptor', () => {
+    function fakeDesc() {}
+    descEvent.event = false;
```
andreybelous, line 6: cosmetic: redundant definition

andreybelous (approved):

src/modules/esl-event-listener/core/descriptors.ts
ala-n: Totally agree. Moved to release 4.3.0 enhancements

Natalie-Smirnova (approved):

julia-murashko (approved):

PR #1440, opened 2023-01-25
fix(esl-image): attribute callback before connect

### LINKED ISSUE #1359: [🐛 in esl-image]: params don't change  when esl-image isn't presented in DOM
**Describe the 🐛 bug in component**
Params don't change  when esl-image isn't presented in DOM

**To Reproduce**
Examples with dynamic modal 

**Current behavior**
params don't change

**Expected behavior**
params should change

fshovchko:
In the scope of this pull request i fixed the issue, where some of the parameters did not change, if were modified before esl-image connected callback fire.
It was found out that it was happening only with `mode` attribute. On of the reasons of such behavior was because connectedCallback doesn't know the previous value of the attribute, so the change can be captured in `attributeChangeCallback`.
But it was later discovered that copied esl-image was losing it's context, the `this._strategy` property. In order to fix this, i propose to 'restore' context by modifying `syncImage` method and passing it old value of attribute from `attributeChangeCallback`.
There are other ways of fixing the issue, for example modifying the strategies `clear` method, so it would be single for every mode strategy. But at current stage, the solution works the best
Closes #1359


src/modules/esl-image/core/esl-image.ts
```diff
@@ -89,7 +89,14 @@ export class ESLImage extends ESLBaseElement {
   }
 
   protected attributeChangedCallback(attrName: string, oldVal: string, newVal: string): void {
-    if (!this.connected || oldVal === newVal) return;
+    if (oldVal === newVal) return;
+    if (attrName === 'mode') {
+      this.syncImage(oldVal);
```
ala-n, line 94: It sounds like not a good idea to initialize the old strategy. We should use browser resources for actions that make the actual state of the component.

I see that there may be context loss in the case of cloning node, but it sounds like custom elements should give us logical and definite steps to make a new instance alive (re-init it). Or we can choose some more stable means like lazy getters.

The main problem right now (to provide more visibility to what I'm trying to say) is that we have extra operations in the normal cases, and we have it in a place that is not related to the issue directly.

ala-n (commented):
Let's split the flows that lead to the issue: 
1) The scenario that is not directly related to the original issue - `cloneNode` 
Here as I mention in a comment, we need to investigate the flow technically in theory and choose the means that will provide stability for the whole situation

2) The main scenario mentioned in the issue:
The update of the mode or source of a detached element that had already been initialized has no effect.
Test case:
const image = ESLImage.create();
image.mode = 'origin';
image.src = 'image1.png'
document.body.append(image);
// ~ await
document.body.remove(image);
// now the image is initialized but detached
image.mode = 'fit';
image.src = 'image2.png'
document.body.append(image); // And here we should have a problem with mode originally and with source (still as I see)

src/modules/esl-image/core/esl-image.ts
fshovchko: The problem breaks into two parts issue with cloning element and issue with re-appending.
Cloning element:
-Strategy context lost, meaning previous rules sometimes can't be cleared and new ones added to it. Fix for it is pushed is currently pushed in the branch
Re-appending element:
- Property `this._srcRules` has image source srtored in it. And because we are not capturing the attribute change before connected state, `this._srcRules` doesn't update, so it gets stuck with previous image. For example can be fixed with this code snippet: `this.srcRules = ESLMediaRuleList.parse(this.src);` in connectedCallback

ala-n (approved):
Pre-checked and accepted on the test env

NastaLeo (approved):

yadamskaya (approved):

Natalie-Smirnova (approved):

abarmina (approved):

dshovchko (approved):

PR #1493, opened 2023-02-20
chore(lint): install `html lint` to lint `njk` and `html` files, remo…

### LINKED ISSUE #1409: [🔧 linting]: consider replacing of  the lint-space and html linter integration (require njk support)
As an ESL contributor, I want stable and correct HTML linting. 
Unfortunately, as for now, there is a couple of related problems: 
 - lint-space behaves incorrectly for .njk files right now as well as it is not fully compatible with beautified HTML 
   It is requested to investigate and consider of lint-space tool removal and replacement by other tools
- consider HTML liner integration

Natalie-Smirnova:
Closes: #1409


pages/views/search.njk
```diff
@@ -6,6 +6,6 @@ eleventyExcludeFromCollections: true
 collectionIcon: sidebar/search.svg
 ---
 <esl-d-search-page-wrapper>
-  <div class="animation-loading bg-blue m-auto"></div>
-  <div class="gcse-search" data-resultsUrl="{{ '/search.html' | url }}"></div>
+   <div class="animation-loading bg-blue m-auto"></div>
```
ala-n, line 9: As far as it is looked, it is incorrect there should be a two spaces instead of 3

pages/views/examples/toggleable.njk
```diff
@@ -6,119 +6,119 @@ name: Toggleable
...
-    - esl-trigger
-  examples:
-    - accordion
-    - tabs
+components:
+- esl-toggleable
+- esl-trigger
+examples:
+- accordion
+- tabs
```
ala-n, line 14: That's will not work.
Frontmetter it is yml syntax, so each indent matters.
So it is not right configuration here right now

ala-n (changes requested):

PR #1536, opened 2023-03-10
feat(esl-core): make this.constructor strictly typed for `ESLBaseElement` and `ESLMixinElement`

ala-n:


fshovchko (approved):

NastaLeo (approved):

ala-n:
Reverted due to TS issue with Partial of hacked object
Hacks with generic functions also do not work due to TS not resolving instance type deeper than the default definition.
So according to https://github.com/Microsoft/TypeScript/issues/3841#issuecomment-337560146
unfortunately, we have no more legal options to make it strictly typed as expected.
The shortcut for tagName is implemented in #1551, and that's all until a fix from TS appears (if it appears).

Issue #1540, opened 2023-03-10
[🚀esl-utils]: Proposal: `getEventListeners()` method for `SynteticEventTarget`


ala-n:
As an ESL consumer, I want to have a more understandable check for a listener's count than `this.hasEventListener(type, min)`

Proposal: 
Add `getEventListeners` method with the following signature:
```typescript
   protected getEventListeners(): EventListener[]; // Returns all event listeners of the SyntetycEvent target
   protected getEventListeners(type: string): EventListener[]; // Returns all event listeners of the SyntetycEvent target that handles passed event type
```

NTH: reuse method inside `hasEventListener()`
Deprecate: numeric and second parm for `hasEventListener()`

Internal code update and deprecation suggestion:
`this.hasEventListener(event, 1)` => `this.getEventListeners(event).length > 1`

PR #1548, opened 2023-03-14
feat(esl-utils): getEventListeners() method for SynteticEventTarget

fshovchko:
Protected method to get event listeners for SynteticEventTarget
#1540

### DIFF
--- a/src/modules/esl-utils/dom/events/target.ts
+++ b/src/modules/esl-utils/dom/events/target.ts
@@ -9,12 +9,23 @@ export class SyntheticEventTarget implements EventTarget {
 
   private readonly _listeners: Record<string, EventListenerOrEventListenerObject[]> = {};
 
+  protected getEventListeners(): EventListenerOrEventListenerObject[];
+  protected getEventListeners(type: string): EventListenerOrEventListenerObject[];
+  protected getEventListeners(type?: string): EventListenerOrEventListenerObject[] {
+    if (!type) return Object.values(this._listeners).flat(1);
+    if (!this._listeners[type]) Object.assign(this._listeners, {[type]: []});
+    return this._listeners[type];
+  }
+
   public hasEventListener(): boolean;
+  public hasEventListener(type: string): boolean;
+  /** @deprecated alias for `addEventListener` */
   public hasEventListener(type: string | number): boolean;
+  /** @deprecated alias for `addEventListener` */
   public hasEventListener(type: string, minCount: number): boolean;
   public hasEventListener(type: string | number = 'change', minCount: number = 0): boolean {
-    if (typeof type !== 'string') return this.hasEventListener('change', type || 0);
-    return this._listeners[type]?.length > minCount;
+    if (typeof type !== 'string') return this.hasEventListener('change', type || 0); // TODO: remove in 5.0.0
+    return this.getEventListeners(type).length > minCount;
   }
 
   public addEventListener(callback: EventListenerOrEventListenerObject): void;
@@ -23,9 +34,9 @@ export class SyntheticEventTarget implements EventTarget {
     if (typeof type !== 'string') return this.addEventListener((this.constructor as typeof SyntheticEventTarget).DEFAULT_EVENT, type);
 
     validateEventListenerType(callback);
-    if (this._listeners[type] && this._listeners[type].includes(callback!)) return;
-    if (this._listeners[type]) this._listeners[type].push(callback!);
-    else Object.assign(this._listeners, {[type]: [callback]});
+    const listeners = this.getEventListeners(type);
+    if (listeners.includes(callback!)) return;
+    listeners.push(callback!);
   }
 
   public removeEventListener(callback: EventListenerOrEventListenerObject): void;
@@ -34,8 +45,9 @@ export class SyntheticEventTarget implements EventTarget {
     if (typeof type !== 'string') return this.removeEventListener((this.constructor as typeof SyntheticEventTarget).DEFAULT_EVENT, type);
 
     validateEventListenerType(callback);
-    if (!this._listeners[type]) return;
-    this._listeners[type] = this._listeners[type].filter((cb) => cb !== callback);
+    const listeners = this.getEventListeners(type);
+    const index = listeners.indexOf(callback!);
+    if (index >= 0) listeners.splice(index, 1);
   }
 
   public dispatchEvent(e: Event, target: EventTarget = this): boolean {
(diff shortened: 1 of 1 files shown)


src/modules/esl-utils/dom/events/target.ts
```diff
@@ -9,12 +9,23 @@ export class SyntheticEventTarget implements EventTarget {
 
   private readonly _listeners: Record<string, EventListenerOrEventListenerObject[]> = {};
 
+  protected getEventListeners(): EventListenerOrEventListenerObject[];
+  protected getEventListeners(type: string): EventListenerOrEventListenerObject[];
+  protected getEventListeners(type?: string): EventListenerOrEventListenerObject[] {
+    if (!type) return Object.values(this._listeners).flat(1);
+    if (!this._listeners[type]) Object.assign(this._listeners, {[type]: []});
```
ala-n, line 16:
```suggestion
    if (!this._listeners[type]) return [];
```
There is no need to mutate state, this getter should not have side effects

src/modules/esl-utils/dom/events/target.ts
```diff
@@ -34,8 +45,9 @@ export class SyntheticEventTarget implements EventTarget {
     if (typeof type !== 'string') return this.removeEventListener((this.constructor as typeof SyntheticEventTarget).DEFAULT_EVENT, type);
 
     validateEventListenerType(callback);
-    if (!this._listeners[type]) return;
-    this._listeners[type] = this._listeners[type].filter((cb) => cb !== callback);
+    const listeners = this.getEventListeners(type);
+    const index = listeners.indexOf(callback!);
+    if (index >= 0) listeners.splice(index, 1);
```
ala-n, line 50: The previous version of the code is considered as more stable because it assumes that array can be corrupted by not unique values

src/modules/esl-utils/dom/events/target.ts
```diff
@@ -48,7 +60,7 @@ export class SyntheticEventTarget implements EventTarget {
     Object.defineProperty(e, 'currentTarget', targetDescriptor);
     Object.defineProperty(e, 'srcElement', targetDescriptor);
 
-    this._listeners[e.type]?.forEach((listener) => {
+    this.getEventListeners(e.type).forEach((listener) => {
```
ala-n, line 63: Then we should make a check for e.type
Because if for some reason we pass an event with an empty type we will get totally incorrect dispatching behavior

src/modules/esl-utils/dom/events/target.ts
```diff
@@ -9,12 +9,23 @@ export class SyntheticEventTarget implements EventTarget {
 
   private readonly _listeners: Record<string, EventListenerOrEventListenerObject[]> = {};
 
+  protected getEventListeners(): EventListenerOrEventListenerObject[];
+  protected getEventListeners(type: string): EventListenerOrEventListenerObject[];
+  protected getEventListeners(type?: string): EventListenerOrEventListenerObject[] {
+    if (!type) return Object.values(this._listeners).flat(1);
```
ala-n, line 15: 1) the collection that we return should be unique
2) as this is not beta and the current target is 4, we can not overuse ES6 built-in methods as officially we still have ie11 until 5th version is released (We can assume Object.values will be covered by polyfill, but as for now flat 
 is better to be consumed from esl-utils/misc/array)

ala-n (changes requested):

dshovchko (approved):

NastaLeo (approved):

yadamskaya (approved):

ala-n:
Hey, @exadel-inc/esl-all-contributors  team the is a small stylistic question appears in this PR, so we need your feedback to decide the final way for our update. 
The question relates to the uniqueness property in the `getEventListener` tech implementation.
We need it from the API point of view independently of passing the actual event or getting all listeners.
But technically, we have some details.

Option 1 (👍🏻) 
Currently in PR
For consistency, we have used the `uniq` for parametrized and general flow. 
```
protected getEventListeners(type?: string): EventListenerOrEventListenerObject[] {
    return uniq((typeof type !== 'string') ? flat(Object.values(this._listeners)) : (this._listeners[type] || []));
}
```

At the same time, we do not use `getEventListeners(type)` inside the `dispatchEvent` as it is a significant performance loss
`dispatchEvent` can be called frequently, so creating a Set under the hood is not a good idea
 ```
 // dispatchEvent part
    this._listeners[e.type]?.forEach((listener) => {
      if (typeof listener === 'function') listener.call(this, e);
      else listener.handleEvent.call(listener, e);
    });
 ```
 
 But potentially, we share overoptimized getEventListeners with the user. 
 (But just to be clear, this is an edge case that user access listeners and do it so frequently)
 
 Option 2 (🎉)
 We do not make extra operations inside the `getEventListeners`
 ```
protected getEventListeners(type?: string): EventListenerOrEventListenerObject[] {
    if (typeof type !== 'string') return uniq(flat(Object.values(this._listeners)));
    return this._listeners[type] || [];
}
 ```
 The collections (arrays) in the private store `this._listeners` are unique separately as they are modified by `addEventListener / removeEventListener` accordingly (saving uniq state)
 
Then we can safely use it in dispatchEvent
 ```
 // dispatchEvent part
    this.getEventListeners(e.type).forEach((listener) => {
      if (typeof listener === 'function') listener.call(this, e);
      else listener.handleEvent.call(listener, e);
    });
 ```
 
 But again, potentially, the key `this._listeners` are still accessible and can contain invalid data (edge case too)

As additional options, we can also consider
Option 3 (👀)  = Option 2 + store listeners under the true private symbol (the lightweight enough change, but do we need it)

Option 4 (🚀) use a Set to store listeners (but according to the specific of the event target, the only frequent operation is getting listeners for a defined event type, so there is may no need to use mo strict collection)

Please share your opinion with marked emoji or in the tread (here or Slack)

Natalie-Smirnova (approved):

ala-n:
According to the votes, let's go with the following update for the `getEventListener` tech implementation
1) No extra processing on the internal collection
```
protected getEventListeners(type?: string): EventListenerOrEventListenerObject[] {
   if (typeof type !== 'string') return uniq(flat(Object.values(this._listeners)));
   return this._listeners[type] || [];
}
```
2) Simplify dispatch
```
// dispatchEvent part
   this.getEventListeners(e.type).forEach((listener) => {
     if (typeof listener === 'function') listener.call(this, e);
     else listener.handleEvent.call(listener, e);
   });
```
3) Make `this._listeners` stored under the hidden Symbol key (true private implementation)

src/modules/esl-event-listener/core/targets/decorated.target.ts
```diff
@@ -51,7 +51,7 @@ export class ESLEventTargetDecorator<Args extends any[]> extends SyntheticEventT
   public override addEventListener(event: any, callback: EventListener = event): void {
     super.addEventListener(event, callback);
 
-    if (!this.hasEventListener(event, 1)) {
+    if (this.getEventListeners(event).length < 2) {
```
ala-n, line 54:
```suggestion
    if (this.getEventListeners(event).length > 1) return;
```
Let's implement the check consistent across the places

NastaLeo (approved):

Natalie-Smirnova (approved):

ala-n (approved):

Issue #1562, opened 2023-03-17
[🚀esl-event-listener]: Support for `ESLDomElementTarget` as a target for `ESLResizeObserverTarget`


ala-n:
As an ESL consumer, I want to be able to pass a `ESLDomElementTarget` instance to ESLEventUtils.resize to have consistent shortcuts for both CustomElements and `ESLMixinElement`s

Note: consider rethinking the criteria for `ESLDomElementTarget` detection. An HTMLElement inheritor can not be considered as `ESLDomElementTarget` itself

PR #1573, opened 2023-03-27
feat(esl-utils): Support for ESLDomElementTarget as a target for ESLResizeObserverTarget

fshovchko:
Target for ResizeObserverTarget with extended from 'Element' to being able to accept mixin elements.  #1562


src/modules/esl-event-listener/core/targets/resize.adapter.ts
```diff
@@ -10,6 +15,14 @@ export class ESLResizeObserverTarget extends SyntheticEventTarget {
     changes.forEach(this.handleChange, this)
   );
 
+  public readonly target: Element;
+
+  private static getActualTarget(target: ESLResizeTarget): Element {
+    if (target instanceof Element) return target;
+    const host = getPropertyDescriptor(target, '$host') as Element;
```
ala-n, line 22: How does the property descriptor relate to the host value ? 
Is it not compatible;e with Element, and it is not something that the ESLResizeObserverTarget observer should use.

ala-n (changes requested):

src/modules/esl-event-listener/core/targets/resize.adapter.ts
```diff
@@ -1,47 +1,54 @@
...
-   * Note the {@link ESLResizeObserverTarget} instances are singletons relatively to the {@link Element}
+   * Creates {@link ESLResizeObserverTarget} for the {@link ESLResizeTarget}.
+   * Note the {@link ESLResizeObserverTarget} instances are singletons relatively to the {@link ESLResizeTarget}
    */
-  protected constructor(
-    /** Observed {@link Element} of the {@link ESLResizeObserverTarget} instance */
-    public readonly target: Element
-  ) {
+  protected constructor(target: ESLResizeTarget) {
+    if (!(target instanceof Window || target instanceof Element)) target = target.$host;
```
ala-n, line 43: Consider mowing to esl-utils/abstract/dom-target as utility `resolveDomTarget`
That can also be applied to other cases of ESLDomElementTarget usage

src/modules/esl-event-listener/core/targets/resize.adapter.ts
```diff
@@ -1,47 +1,54 @@
 import {SyntheticEventTarget} from '../../../esl-utils/dom/events/target';
 import {overrideEvent} from '../../../esl-utils/dom/events/misc';
 
+import type {ESLDomElementTarget} from '../../../esl-utils/abstract/dom-target';
+
+export type ESLResizeTarget = ESLDomElementTarget | Window;
```
ala-n, line 6: Consider removing it due to only one actual usage, and no practical need for users as well

ala-n (changes requested):
This was not an exception to creating tasks regarding ESLElementRelated and window separately.

The first one relates to creating `resolveDomTarget` basically, and it will most likely be reused 
It will be a simple change for ESLResizeObserverTarget but it relates to 
 - discussion of the `resolveDomTarget` function usage, name and etc.
 
 The second issue regarding window is a completely different story, we need to think about unified API. And it also has a chance to be even rejected.

ala-n (approved):

NastaLeo (approved):

Issue #1535, opened 2023-03-09
[🐛 in esl-utils]:  `throttle` looks to be handled incorrectly


ala-n:
### Steps:
1. 
```
const onRes = throttle((e) => console.log('Throttled catch: ', e), 50);
window.addEventListener('resize', throttle);
```
2. Minimize/maximize the window
3. Observe watched events

### Current:
Throttled event handler called 3 times

### Expected:
Throttled event handler called less than 3 times

PS: window.resize happens multiple times during the "Minimize/maximize window" action -> expected number of calls is 2

PR #1595, opened 2023-04-06
fix(esl-utils): incorrect throttle behaviour

fshovchko:
Throttle looked to be handled incorrectly
#1535

### DIFF
--- a/src/modules/esl-event-listener/test/targets/decorator.test.ts
+++ b/src/modules/esl-event-listener/test/targets/decorator.test.ts
@@ -141,6 +141,7 @@ describe('ESLEventUtils.decorate proxy', () => {
       window.dispatchEvent(event);
       jest.advanceTimersByTime(10);
       expect(fn).lastCalledWith(event);
+      jest.advanceTimersByTime(DEFAULT_TIMEOUT + 100);
     });
 
     test('Proxy scroll happens throttled (multiple events received once in bounds of threshold)',  ()=> {
--- a/src/modules/esl-event-listener/test/targets/decorator.test.ts
+++ b/src/modules/esl-event-listener/test/targets/decorator.test.ts
@@ -124,29 +124,31 @@ describe('ESLEventUtils.decorate proxy', () => {
     const host = {};
     const fn = jest.fn();
 
-    beforeAll(() => {
+    beforeEach(() => {
       ESLEventUtils.subscribe(host, {
         event: 'scroll',
         target: ESLEventUtils.decorate(window, throttle, DEFAULT_TIMEOUT)
       }, fn);
       jest.useFakeTimers();
+      fn.mockReset();
     });
-    afterAll(() => ESLEventUtils.unsubscribe(host));
 
-    beforeEach(() => fn.mockReset());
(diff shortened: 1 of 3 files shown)


src/modules/esl-event-listener/test/targets/decorator.test.ts
```diff
@@ -141,6 +141,7 @@ describe('ESLEventUtils.decorate proxy', () => {
       window.dispatchEvent(event);
       jest.advanceTimersByTime(10);
       expect(fn).lastCalledWith(event);
+      jest.advanceTimersByTime(DEFAULT_TIMEOUT + 100);
```
ala-n, line 144: ??? Why do we need it? we have a reset before each test
```
beforeEach(() => fn.mockReset());
afterEach(() => jest.runAllTimers());
```

src/modules/esl-utils/async/throttle.ts
```diff
@@ -26,21 +26,21 @@ export function throttle<F extends AnyToAnyFnSignature>(fn: F, threshold = 250,
 
   function throttledSubject(...args: any[]): void {
     const now = Date.now();
+    const lastThreshold = last + threshold;
 
-    if (!last || now >= last + threshold) {
-      last = now;
-      fn.apply(this, args);
-    }
+    deferred = createDeferred();
```
ala-n, line 31: The behavior of the result promise changed.
First of all, it's changed incorrectly: as for right now, it will produce promises that never be resolved (memory leak).
Second thing, not sure about the deferred and result promise in general for `throttle`.
We need to have a logical principal for result promise resolution, or we need to consider getting rid of that feature at all for `throttle`

ala-n (changes requested):

dshovchko (approved):

src/modules/esl-utils/async/throttle.ts
```diff
@@ -26,24 +26,20 @@ export function throttle<F extends AnyToAnyFnSignature>(fn: F, threshold = 250,
...
     (typeof timeout === 'number') && clearTimeout(timeout);
     timeout = window.setTimeout(() => {
-      last = now;
+      last = Date.now();
       timeout = null;
-      // fn.apply to save call context
       deferred!.resolve(fn.apply(thisArg || this, args));
       deferred = null;
-    }, threshold);
+    }, (last || now < lastThreshold) ? lastThreshold - now : 0);
```
ala-n, line 39: The check `(last || now < lastThreshold)` looks a little bit complicated
So the only moment when it is not true is the initial call when `last` is not defined
but in that case then `now < lastThreashold` ~ ` now < undefined + treshold ` which is not true starting from at least 2th of January 1970

Other words it sounds like 
`(last || now < lastThreshold) ? lastThreshold - now : 0`
->
`(now < lastThreshold) ? lastThreshold - now : 0` (which is super logical and there is no questions for such check)
->
`Math.max(lastThreshold - now, 0)`

ala-n (approved):

NastaLeo (approved):

ala-n (approved):

PR #1602, opened 2023-04-11
refactor(esl-event-listener): change API of extended EventTargets

ala-n:
BREAKING CHANGE: beta `ESLEventUtils.resize` replaced with `ESLResizeObserverTarget.for`
BREAKING CHANGE: beta `ESLEventUtils.decorate` replaced with `ESLDecoratedEventTarget.for`


src/modules/esl-event-listener/core/targets/resize.adapter.ts
```diff
@@ -1,6 +1,8 @@
 import {SyntheticEventTarget} from '../../../esl-utils/dom/events/target';
 import {ESLElementResizeEvent} from './resize.adapter.event';
 
+export {ESLElementResizeEvent};
```
NastaLeo, line 4: do we really need this export?

NastaLeo (approved):

src/modules/esl-event-listener/core/targets/resize.adapter.ts
ala-n: Yep, it is functional.
Otherwise, we will need to mention the events file in the module cumulative file, but we can consider this option as well
Do you voting for moving export to `src/modules/esl-event-listener/core.ts` ?

abarmina (approved):

Natalie-Smirnova (approved):

PR #1668, opened 2023-05-22
esl-share: create popup mode

dshovchko:
In the scope of PR:
 - added `mode="popup"` for esl-share
 - created `esl-share-trigger` component that forwarded share attributes from parent esl-share to trigger target on open
 - the logic of determining attribute values for sharing (title to share and url to share) has been changed
 - demo pages reworked


yadamskaya (approved):

julia-murashko (approved):

pages/views/draft/share-regression.njk
```diff
@@ -0,0 +1,132 @@
...
+  <p>Shows all available buttons from config.</p>
+  <esl-share list="all"></esl-share>
+  <hr>
+
+  <h3>Displays list of share buttons with empty list (list="")</h3>
+  <p>Shows empty component without buttons.</p>
+  <esl-share list=""></esl-share>
+  <hr>
+
+  <h3>Displays list of share buttons on nonexistent group</h3>
```
fshovchko, line 51:
```suggestion
  <h3>Displays list of share buttons for nonexistent group</h3>
```

pages/views/draft/share-regression.njk
```diff
@@ -0,0 +1,132 @@
...
+  <p>Shows empty list when a nonexistent group is specified.</p>
+  <esl-share list="group:foo"></esl-share>
+  <hr>
+
+  <h3>Displays list of share buttons only existent groups or buttons</h3>
```
fshovchko, line 56:
```suggestion
  <h3>Displays list of share buttons only for existent groups or buttons</h3>
```

src/modules/esl-share/README.md
```diff
@@ -128,6 +128,12 @@ ESLShare.config(() => fetch('/assets/share/config.json').then((response) => resp
  - `additional` - additional params to pass into a button (can be used by share actions)
  - `unavailable` - marker of availability of share button
 
+#### Attributes cascading
+
+If you need to use URL and title overrides within any parts of the page to share using the `share-url` and `share-title` attributes, you don't need to write these attributes on each button. You can write them once on the root element of the part for which these values are valid. You can do this on the body of the document if you want to override the values for the entire document.
```
fshovchko, line 133:
```suggestion
If you want to utilize URL and title overrides within any parts of the page to share using the `share-url` and `share-title` attributes, there's no need to write these attributes on each button. You can write them once on the root element of the part for which these values are valid. Alternatively, if you want to override the values for the entire document, you can set them on the body of the HTML document.
```

src/modules/esl-share/README.md
```diff
@@ -139,7 +145,17 @@ ESLShare.config(() => fetch('/assets/share/config.json').then((response) => resp
...
+ - `mode` - rendering mode of the share buttons. The `list` and `popup` are available (list by default)
  - `ready` - ready state marker
+
+#### Modes
+
+There are two modes available to render buttons.
+
+In `list` mode, the buttons are drawn inside the component as a list. Nothing special.
+
+When `popup` mode is specified, the buttons are created inside a [ESLPopup](../esl-popup/README.md) element, which is built directly into the document's body. If a [ESLPopup](../esl-popup/README.md) element with the desired set of buttons already exists in the document body, the existing one will be reused. A trigger element is created inside the ESLShare component to activate the popup with share buttons, which will activate the popup when you hover over it. Also, one additional activity of the ESLShareTrigger is to forward the `share-title` and `share-url` attributes from the root ESLShare component to the popup. So it's possible for components with the same set of buttons but different URLs and title to share to use the same popup.
```
fshovchko, line 157:
```suggestion
When `popup` mode is specified, the buttons are created inside of a [ESLPopup](../esl-popup/README.md) element, which is built directly into the document's body. If a [ESLPopup](../esl-popup/README.md) element with the desired set of buttons already exists in the document body, the existing one will be reused. A trigger element is created inside the ESLShare component to activate the popup with share buttons, which will activate the popup when you hover over it. Also, one additional activity of the ESLShareTrigger is to forward the `share-title` and `share-url` attributes from the root ESLShare component to the popup. So it's possible for components with the same set of buttons but different URLs and title to share to use the same popup.
```

fshovchko, line 157:
```suggestion
When `popup` mode is specified, the buttons are created inside of a [ESLPopup](../esl-popup/README.md) element, which is built directly into the document's body. If a [ESLPopup](../esl-popup/README.md) element with the desired set of buttons already exists in the document body, the existing one will be reused. A trigger element is created inside the ESLShare component to activate the popup with share buttons, which will activate the popup when you hover over it. Also, one additional activity of the ESLShareTrigger is to forward the `share-title` and `share-url` attributes from the root ESLShare component to the popup. So it's possible for components with the same set of buttons but different URLs and title to share to use the same popup.
```

abarmina (approved):
very cool!

ala-n (approved):
Approved for the epic branch.
And it's definitely fine to be released on the beta channel.

But still not sure about the architecture, or the responsibilities division in the module, to be particular.
I see a couple of details that make me think that my concerns may appear to be correct:
1. First of all, it's `ESLShareTrigger`. From my perspective, it 
looks totally fine from the user API.
But the fact that it does such small and abstract things in combination with the actual behavior (it is totally internal for now as it does not do anything in case the popup is not created and passed) sounds like something went wrong.
2. The ESLShare root component is not so bad at all. Also, it is expected that some parts of such a complicated module may appear to be not completely single-responsible. But it looks like we have an opportunity to make the ESLShare root component less overloaded.

I see the resolution of the mentioned problems in the following directions:
   - Moving the popup instance management to ESLShareTrigger.
It sounds like if it is the ESLShareTrigger zone of responsibility, as a result we will have a much more clear, stable, and logically separated implementation
   - Split away the config-to-button transformation from  ESLShare. This part will block us from making the first transition smooth. And it is also one of the things that looks to be not a zone of responsibility for the root ESLShare module.
   
Let's discuss it today on our weekly sink-up, mb we can shift the stable release a little bit to make this architecture closer to ideal.

NastaLeo (approved):

src/modules/esl-share/core/esl-share.ts
```diff
@@ -70,17 +20,37 @@ function getButtonsList(config: ShareConfig, list: string): ShareButtonConfig[]
...
   /** Event to dispatch on ready state of {@link ESLShare} */
   @prop('esl:share:ready') public SHARE_READY_EVENT: string;
 
+  /** Default initial params to pass into the newly created popup */
+  @prop({
+    position: 'top',
+    defaultParams: {
+      hideDelay: 200
+    }
+  }) public popupInitialParams: PopupActionParams;
```
ala-n, line 52:
```suggestion
  }) protected popupInitialParams: PopupActionParams;
```
Due to the follow-up reasons

ala-n (approved):

PR #1766, opened 2023-07-03
fix(esl-core): register checks improvements

ala-n:


julia-murashko (approved):

src/modules/esl-base-element/core/esl-base-element.ts
```diff
@@ -108,16 +108,16 @@ export abstract class ESLBaseElement extends HTMLElement implements ESLBaseCompo
    */
   public static register(this: typeof ESLBaseElement, tagName?: string): void {
     tagName = tagName || this.is;
-    if (!tagName) throw new Error('Can not define custom element');
-    const constructor: any = customElements.get(tagName);
-    if (constructor) {
-      if (constructor.is !== tagName) throw new Error('Element declaration tag inconsistency');
-      return;
+    if (!tagName) {
+      throw new DOMException('[ESL]: Incorrect tag name', 'NotSupportedError');
```
dshovchko, line 112: Maybe a should be implemented message formatting method for consistency purposes?

dshovchko (approved):

src/modules/esl-base-element/test/element.test.ts
```diff
@@ -1,5 +1,6 @@
 import '../../../polyfills/es5-target-shim';
 import {ESLBaseElement} from '../core';
+import {raw} from 'concurrently/dist/src/defaults';
```
abarmina, line 3: as I see you don't use this import

src/modules/esl-base-element/test/element.test.ts
```diff
@@ -28,17 +29,15 @@ describe('ESLBaseElement', () => {
...
-      .then(() => {
-        expect(() => TestElement2.register('test-test')).not.toThrowError();
-        TestElement2.is = 'test-test-2';
-        // Tag inconsistency
-        expect(() => TestElement2.register('test-test')).toThrowError();
-      });
+    await customElements.whenDefined('test-test');
+    expect(() => TestElement2.register('test-test')).not.toThrowError();
+    expect(() => TestElement2.register('test-test-2')).toThrowError();
+    try {TestElement2.is = 'test-test-2';} catch {}
```
abarmina, line 39: shall we add something to ```catch```?

abarmina (approved):

src/modules/esl-base-element/test/element.test.ts
ala-n: oh **, missed that
Thanks a lot Ann ❤️

src/modules/esl-base-element/test/element.test.ts
ala-n: I think not.
I'd make an expectation statement here but there is an interesting thing: Node.js throws an error when you try to write immutable property, while browsers just ignore it...

src/modules/esl-base-element/core/esl-base-element.ts
ala-n: Agree I will create a technical cleanup regarding the unification of exception handling

Issue #1799, opened 2023-07-14
[🐛 in esl-toggleable]: incorrect behavior when open attribute is being changed


fshovchko:
Steps to reproduce:
1. Have the toggleable with default behavior, with initially closed state (`this.open=false`).
2. Set the `open` attribute.

It will invoke `attributeChangedCallback`, where the code will pass `if (this.open === this.hasAttribute('open')) return;` check, since the attribute has now a truthy value.  But `this.open` state still equal to `false`, since it's not directly related to the `open` attribute. Then it will call the toggle method with essentially this value: `this.toggle(false, {initiator: 'attribute', showDelay: 0, hideDelay: 0});`. What code is attempting, is to switch between closed state to a closed state

PR #1800, opened 2023-07-14
fix(esl-togglable): open attribute change behavior

fshovchko:
The scope of this PR is to resolve issue described in #1799 + updated error message for toggleable dispatcher. ESLint didn't like the extra declaration, so another way of fixing it may be:
```
    if (!this.connected || newVal === oldVal) return;
    const isOpen = this.hasAttribute('open');
    switch (attrName) {
      case 'open':
        if (this.open === isOpen) return;
        this.toggle(isOpen, {initiator: 'attribute', showDelay: 0, hideDelay: 0});
        break;
```

### DIFF
--- a/src/modules/esl-toggleable/core/esl-toggleable.ts
+++ b/src/modules/esl-toggleable/core/esl-toggleable.ts
@@ -148,12 +148,12 @@ export class ESLToggleable extends ESLBaseElement {
   protected override attributeChangedCallback(attrName: string, oldVal: string, newVal: string): void {
     if (!this.connected || newVal === oldVal) return;
     switch (attrName) {
-      case 'open':
-        // eslint-disable-next-line no-case-declarations
+      case 'open': {
         const isOpen = this.hasAttribute('open');
         if (this.open === isOpen) return;
         this.toggle(isOpen, {initiator: 'attribute', showDelay: 0, hideDelay: 0});
         break;
+      }
       case 'group':
         this.$$fire(this.GROUP_CHANGED_EVENT, {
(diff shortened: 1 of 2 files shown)


src/modules/esl-toggleable/core/esl-toggleable.ts
```diff
@@ -149,8 +149,10 @@ export class ESLToggleable extends ESLBaseElement {
     if (!this.connected || newVal === oldVal) return;
     switch (attrName) {
       case 'open':
-        if (this.open === this.hasAttribute('open')) return;
-        this.toggle(this.open, {initiator: 'attribute', showDelay: 0, hideDelay: 0});
+        // eslint-disable-next-line no-case-declarations
+        const isOpen = this.hasAttribute('open');
+        if (this.open === isOpen) return;
+        this.toggle(isOpen, {initiator: 'attribute', showDelay: 0, hideDelay: 0});
```
ala-n, line 155:
``` suggestion
      case 'open': {
        const isOpen = this.hasAttribute('open');
        if (this.open === isOpen) return;
        this.toggle(isOpen, {initiator: 'attribute', showDelay: 0, hideDelay: 0});
      }
```
Please don't disable the linting rules. 
In case lint have something to say, most likely we need to fix it, otherwise, we need to discus disabling globally. There are only a few moments when we disable some rules locally.

ala-n (approved):

ala-n (approved):

abarmina (approved):

dshovchko (approved):

Issue #1786, opened 2023-07-10
[⚡tests]: test coverage goal v4.10.0 - 5.0.0


ala-n:
As for v4.10.0 the goal is to increase overall test coverage by up to 50%

Potential scope (splittable, feel free to create separate issues or skip items):

**⚠️ IMPORTANT: _split scope by separate branches/PRs, if the target scope is different of the diff, is big enough_**
**⭐ NOTE: _split tests by files even if they are related to the same submodule if that is reasonable from support perspective_**

### Major:
- [x] [⚡for esl-utils] fix `esl-utils/async/test/promise.raf.ts` test file name 
- [x] [⚡for esl-mixin-element] cover `$$cls`, `$$attr`, `$$fire` methods with unit tests
- [ ] [⚡for esl-trigger] cover `esl-trigger` target to toggleable request delegation with unit tests
   _Note: please make sure you create unit tests encapsulated enough - use fake ESLToggleable implementation, use fake events, and make cases clean, simple, and separated_
- [ ] [⚡for esl-trigger] cover `esl-trigger` toggleable tracking functionality with unit tests
   _Note: please make sure you create unit tests encapsulated enough - use fake ESLToggleable implementation, use fake events, and make cases clean, simple, and separated_
- [x] [⚡for esl-toggleable] cover `esl-toggleable` basic cases with unit tests
- [x] [⚡for esl-toggleable] cover `esl-toggleable-dispatcher` with unit tests

### Normal:
- [x] [⚡for esl-toggleable] unit tests for body-class | container-class functionality of `esl-toggleable`
- [x] [⚡for esl-toggleable] unit tests for `request:show|hide` events of `esl-toggleable`
- [ ] [⚡for esl-trigger] cover `esl-trigger` hover behavior with unit tests
   _Note: please make sure you create unit tests encapsulated enough - use fake ESLToggleable implementation, use fake events, and make cases clean, simple, and separated_
- [x] [⚡for esl-toggleable] unit tests for touch hover behavior of `esl-toggleable`
- [x] [⚡for esl-utils] improve poor `esl-utils/dom/attr` tests coverage
- [x] [⚡for esl-toggleable] basic integration tests for `esl-trigger` + `esl-toggleable`

### Minor:
- [x] [⚡for esl-toggleable] unit tests for default a11ty of  `esl-toggleable`
- [ ] [⚡for esl-trigger] cover `esl-trigger` default a11ty with unit tests
[...]

PR #1811, opened 2023-07-18
test(esl-toggleable): cover dispatcher with tests

fshovchko:
This PR is a part of 50% voverage goal https://github.com/exadel-inc/esl/issues/1786

### DIFF
--- a/src/modules/esl-toggleable/test/toggleable.dispatcher.test.ts
+++ b/src/modules/esl-toggleable/test/toggleable.dispatcher.test.ts
@@ -1,21 +1,22 @@
+import {SyntheticEventTarget} from '../../esl-utils/dom/events/target';
 import {ESLToggleableDispatcher, ESLToggleable} from '../core';
-import {ESLEventUtils} from '../../esl-event-listener/core/api';
+
+const createToggleableMock = (): ESLToggleable =>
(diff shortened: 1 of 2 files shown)


src/modules/esl-toggleable/test/toggleable.dispatcher.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  describe('ESLToggleableDispatcher instance', () => {
+    test('should initialize with correct tag name and default root as body', () => {
+      expect($dispatcher).toBeInstanceOf(ESLToggleableDispatcher);
+      expect($dispatcher.root).toBe(document.body);
+    });
+
+    test('shouldn`t initialize on the same level twice', () => {
+      ESLToggleableDispatcher.init();
+      expect(document.querySelectorAll(ESLToggleableDispatcher.is)).toHaveLength(1);
```
ala-n, line 33: Conflicts with a global state which makes this unit test case invalid by violation of isolation

ala-n (commented):
Consider a more isolated way of tasting.
Consider the ESLTogglableDispatcher as a component with an open event-based API.
So please think about creating a mock event and toggleable instances. 
It sounds better to have integration testing simplified to the most important cases only and to be separated from more detailed unit tests of ESLToggleableDispatcher

PR #1812, opened 2023-07-18
test(esl-toggleable): integration tests for esl-trigger + esl-toggleable

fshovchko:
This PR is a part of 50% voverage goal https://github.com/exadel-inc/esl/issues/1786

### DIFF
--- a/src/modules/esl-toggleable/test/toggleable.integration.test.ts
+++ b/src/modules/esl-toggleable/test/toggleable.integration.test.ts
@@ -4,16 +4,16 @@ import {ESLTrigger} from '../../esl-trigger/core';
 describe('ESLToggleable custom element - integration with ESLTrigger', () => {
   const $trigger = ESLTrigger.create();
   const $toggleable = ESLToggleable.create();
-  $trigger.dispatchEvent = jest.fn();
   $trigger.$target = $toggleable;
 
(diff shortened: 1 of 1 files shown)


src/modules/esl-toggleable/test/toggleable.integration.test.ts
```diff
@@ -0,0 +1,72 @@
+import {ESLToggleable} from '../core';
+import {ESLTrigger} from '../../esl-trigger/core';
+
+describe('ESLToggleable custom element - integration with ESLTrigger', () => {
+  const $trigger = ESLTrigger.create();
+  const $toggleable = ESLToggleable.create();
+  $trigger.dispatchEvent = jest.fn();
+  $trigger.$target = $toggleable;
```
ala-n, line 8: It isn't an integration test...
Integration test in or case: the primary usage flows with exact components instances used for test cases (no mocks in bouns of components related to the tested modules) 
Unit test: we these the module/submodule according to public API; we avoid assumptions regarding implementation, but we also avoid using real components that are not related to the module - we test the module, but we assume everything outside is any module that corresponds to API, and we assume it is already tested.

ala-n (approved):

ala-n (changes requested):
Missclick see the comments above

PR #1839, opened 2023-07-28
fix(esl-utils): debounced unhandled rejection

### LINKED ISSUE #1801: [🐛 in esl-utils]: 'Uncaught in promise' after `debounce.cancel()` call
### Steps:

- `const debounced = debounce(fn, 200);`
- Make sure you do not access and process `debounced.promise`
- `debounced.cancel()`
- Observe the console

**Current result: **
'Uncaught in promise `undefined`' appears in the console due to Promise.reject() creation. 

**Expected behavior**
No errors in the console are expected

fshovchko:
The scope of this PR is to fix the issue described in #1801, and to refactor code from a function to a class

closes #1801


ala-n (changes requested):
Please consider an alternative approach:
 - do not change debounce (most likely, at all). A similar issue may appear in other async decorators
 - Make the `Deferred` utility improved with a lazy `<instanceof Deferred>.promise` instead
 
 The original request to make a class was related to `Defered` but not a denounced.
 
 Please also provide tests to make sure we have no problem (ideally, first of all, make a unit test cases and make sure they are correct but not successful, and then provide a fix)

src/modules/esl-utils/async/test/promise/defered.test.ts
```diff
@@ -1,14 +1,41 @@
...
-  test('Deferred rejects promise when it is rejected', () => {
+
+  test('Deferred rejected promise when it`s rejected and promise', () => {
     const def$$ = createDeferred();
-    def$$.reject(1);
-    return def$$.promise.catch((n) => expect(n).toBe(1));
+    const value = 'test-error';
+    def$$.reject(value);
+    expect(def$$.promise).rejects.toBe(value);
   });
```
ala-n, line 40: Consider the following list of expectation (draft):
```suggestion
  test('Resolve of Deferred produces resolved promise', () => {
    const def$$ = createDeferred();
    def$$.resolve(1);
    return def$$.promise.then((n) => expect(n).toBe(1));
  });
  test('Rejected Deferred produces rejected promise', () => {
    const def$$ = createDeferred();
    def$$.reject(1);
    return def$$.promise.catch((n) => expect(n).toBe(1));
  });

  test('Deferred resolves initially requested promise when resolved', () => {
    const def$$ = createDeferred();
    +def$$.promise;
    def$$.resolve(1);
    return def$$.promise.then((n) => expect(n).toBe(1));
  });
  test('Deferred rejects initially requested promise when rejected', () => {
    const def$$ = createDeferred();
    +def$$.promise;
    def$$.reject(1);
    return def$$.promise.catch((n) => expect(n).toBe(1));
  });

  describe('Rejected Deferred does not leads to uncaught in promise', () => {
    const throwFn = jest.fn((reason) => {throw reason;});
    beforeAll(() => {
      // Avoid memory leak by adding too many listeners
      process.env.LISTENING_TO_UNHANDLED_REJECTION = String(true);
      process.on('unhandledRejection', throwFn);
    });
    afterAll(() => {
      process.off('unhandledRejection', throwFn);
    });
    test('Deferred does not leads to uncaught', async () => {
      const def$$ = createDeferred();
      def$$.reject(1);
      await promisifyTimeout(0);
      expect(throwFn).not.toBeCalled();
    });
  });
```

julia-murashko (approved):

yadamskaya (approved):

HenadzV (approved):

src/modules/esl-utils/async/promise/defered.ts
```diff
@@ -1,23 +1,43 @@
...
-  resolve: (arg: T) => void;
-  /** Function that rejects wrapped promise */
-  reject: (arg?: any) => void;
-};
+import {memoize} from '../../decorators/memoize';
 
-/**
- * Creates Deferred Object that wraps promise and its resolve and reject callbacks
- */
+/** A deferred object represents promise with it's resolve/reject methods */
```
NastaLeo, line 3: why instead of its it's?

NastaLeo (approved):

dshovchko (approved):

Issue #1809, opened 2023-07-18
[🚀 for esl-event-listener]: swipe gesture synthetic event target


ala-n:
As an ESL consumer, I want to have the ability to detect simple touch swipe gestures w/o external dependencies.

Proposed to add `SwipeGestureTarget` utility to the list of built-in event targets.

## API

Basic sample:
```ts
  @listen({
      event: 'swipe'
      target: ESLSwipeGestureTarget.for($target)
  })
  onSwipe(e: ESLSwipeGestureEvent): void {
    // ...
  }
```

Sample with configuration: 
```ts
  @listen({
      event: 'swipe'
      target: ESLSwipeGestureTarget.for($target, {timeout: 200, threshold: '100px'})
  })
  onSwipe(e: ESLSwipeGestureEvent): void {
    // ...
  }
```

Event: 
```ts
interface ESLSwipeGestureEvent {
    /** The gesture direction relatively to viewport */
    direction: 'left' | 'right' | 'up' | 'down';
    /** The distance between start and end point */
    distance: number;
    /** The angle of direction relatively to the top vector (clock-like behaviour) */
    handle: number;
    /** The distance between start and end point along the x axis */
    distanceX: number;
     /** The distance between start and end point along the y axis */
    distanceY: number;
    /** An initial event that caused the swipe gesture */
    startEvent: PointerEvent;
    /** An original event that resolved to initiate a swipe gesture */
    endEvent: PointerEvent;
}
```

Supported event names:
  - `swipe`

Settings API
  ```ts
     interface ESLSwipeGestureSetting {
         /** Timeout threshold */
         timeout: number;
         /**  Support px and vw/vh units */
         threshold: string;  // '10px' or '10vh' //
     }
  ```
 
 The final Public API:
 ```
 /**
  * @param $target - a target element to observe pointer events to detect a gesture 
  * @param config - optional config override (will be merged with a default one if passed)
  */
 ESLSwipeGestureTarget.for($target: HTMLElement, config?: ESLSwipeGestureSetting): EventTarget;
 ```  

### Technical notes:
 - The cache usage is not required, so the decision is on the implementation side.
 
### Scope: 
 - [x] Implement ESLSwipeGestureTarget module and related APIs under `esl-event-listener/core/targets`
 - [x] Provide the test page with a showcase of ESLSwipeGestureTarget
 - [x] Provide short documentation notes under the module README.md file
 - [x] (optional, can be moved to follow up) Provide the module unit tests

[...]

PR #1841, opened 2023-07-31
feat(swipe-target): Add SwipeEventTarget

grechihinrhp:
Add SwipeEventTarget to subscribe `swipe`, `swipe:left`, `swipe:right`, `swipe:up`, `swipe:down` events using `ESLEventUtils.subscribe` or `@listen` descriptor

Closes: #1809

### DIFF
--- a/src/modules/esl-event-listener/core/targets/swipe.event.ts
+++ b/src/modules/esl-event-listener/core/targets/swipe.event.ts
@@ -19,7 +19,7 @@ export interface ESLSwipeGestureEventInfo {
   direction: SwipeDirection;
   /** Distance between the points where pointerdown and pointerup events occurred along the x axis */
   distanceX: number;
-  /** Distance between the points where pointerdown and pointerup events occurred along the x axis */
+  /** Distance between the points where pointerdown and pointerup events occurred along the y axis */
   distanceY: number;
   /** Original pointerdown event */
   startEvent: PointerEvent;
--- a/src/modules/esl-event-listener/core/targets/swipe.event.ts
+++ b/src/modules/esl-event-listener/core/targets/swipe.event.ts
@@ -1,0 +1,47 @@
+import {overrideEvent} from '../../../esl-utils/dom/events/misc';
+
+/**
+ * Swipe directions that could be provided in {@link ESLSwipeGestureEvent}
+ */
+export type SwipeDirection = 'left' | 'right' | 'up' | 'down';
+
+/**
+ * Event names that could be triggered by {@link ESLSwipeGestureTarget}
+ */
+export type SwipeEventName = 'swipe' | 'swipe:left' | 'swipe:right' | 'swipe:up' | 'swipe:down';
+
+/**
+ * Describes swipe information provided with {@link ESLSwipeGestureEvent}
+ */
+export interface ESLSwipeGestureEventInfo {
+  target: Element;
+  /** Swipe direction {@link SwipeDirection} */
+  direction: SwipeDirection;
+  /** Distance between the points where pointerdown and pointerup events occurred along the x axis */
+  distanceX: number;
+  /** Distance between the points where pointerdown and pointerup events occurred along the x axis */
+  distanceY: number;
+  /** Original pointerdown event */
+  startEvent: PointerEvent;
+  /** Original pointerup event */
+  endEvent: PointerEvent;
+}
+
+/**
+ * Creates swipe event dispatched by {@link ESLSwipeGestureTarget}
+ */
+export class ESLSwipeGestureEvent extends UIEvent {
+  public override readonly target: Element;
+  public swipeInfo: ESLSwipeGestureEventInfo;
+
+  protected constructor(eventName: SwipeEventName, swipeInfo: ESLSwipeGestureEventInfo) {
+    super(eventName, {bubbles: true, cancelable: true});
+    this.swipeInfo = swipeInfo;
+    overrideEvent(this, 'target', swipeInfo.target);
+  }
+
+  /** Creates {@link ESLSwipeGestureEvent} from {@link ESLSwipeGestureTarget} */
+  public static fromConfig(eventName: SwipeEventName, swipeInfo: ESLSwipeGestureEventInfo): ESLSwipeGestureEvent {
+    return new ESLSwipeGestureEvent(eventName, swipeInfo);
+  }
+}
--- a/src/modules/esl-event-listener/core/targets/swipe.target.ts
+++ b/src/modules/esl-event-listener/core/targets/swipe.target.ts
@@ -26,15 +26,19 @@ export interface ESLSwipeGestureSetting {
   timeout?: number;
 }
 
+/**
+ * Synthetic target class that produces swipe events
+ */
 export class ESLSwipeGestureTarget extends SyntheticEventTarget {
-  protected startEl: HTMLElement | null;
-  protected startEvent: PointerEvent;
-  protected config: SwipeEventTargetConfig;
-  protected target: Element;
   protected static defaultConfig: SwipeEventTargetConfig = {
     threshold: '20px',
     timeout: 500
   };
+
+  protected startEvent: PointerEvent;
+  protected config: SwipeEventTargetConfig;
+  protected target: Element;
+  protected isGestureStarted: boolean = false;
 
   protected constructor(target: ESLDomElementTarget, settings: ESLSwipeGestureSetting) {
     super();
@@ -59,8 +63,8 @@ export class ESLSwipeGestureTarget extends SyntheticEventTarget {
    */
   @bind
   protected handleStart(e: PointerEvent): void {
-    this.startEl = e.target as HTMLElement;
     this.startEvent = e;
+    this.isGestureStarted = true;
   }
 
   /**
@@ -71,7 +75,7 @@ export class ESLSwipeGestureTarget extends SyntheticEventTarget {
   @bind
   protected handleEnd(e: PointerEvent): void {
     // if the user released on a different target, cancel!
-    if (this.startEl !== e.target) return;
+    if (!this.isGestureStarted || (this.startEvent?.target !== e.target)) return;
     const direction = this.resolveDirection(e);
 
     if (direction) {
@@ -90,8 +94,8 @@ export class ESLSwipeGestureTarget extends SyntheticEventTarget {
       this.dispatchEvent(ESLSwipeGestureEvent.fromConfig(`swipe:${direction}` as SwipeEventName, swipeInfo));
     }
 
-    // reset values
-    this.startEl = null;
+    // Mark swipe as completed
+    this.isGestureStarted = false;
   }
 
   /**
--- a/src/modules/esl-event-listener/core/targets/swipe.target.ts
+++ b/src/modules/esl-event-listener/core/targets/swipe.target.ts
@@ -2,17 +2,20 @@ import {SyntheticEventTarget} from '../../../esl-utils/dom/events/target';
 import {ESLMixinElement} from '../../../esl-mixin-element/ui/esl-mixin-element';
 import {overrideEvent} from '../../../esl-utils/dom/events/misc';
 import {resolveDomTarget} from '../../../esl-utils/abstract/dom-target';
+import {bind} from '../../../esl-utils/decorators/bind';
+import {ESLEventUtils} from '../api';
+
 import type {ESLDomElementTarget} from '../../../esl-utils/abstract/dom-target';
 
 type SwipeDirection = 'left' | 'right' | 'up' | 'down';
 type SwipeEventName = 'swipe' | 'swipe:left' | 'swipe:right' | 'swipe:up' | 'swipe:down';
 
-interface SwipeEvent {
+interface ESLSwipeGestureEvent {
   direction: SwipeDirection;
   distanceX: number;
   distanceY: number;
-  touchStartOriginalEvent: PointerEvent;
-  touchEndOriginalEvent: PointerEvent;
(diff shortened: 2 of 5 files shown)


grechihinrhp:
I have read the CLA Document and I hereby sign the CLA

ala-n:
Thanks a lot for your contribution!
TS Doc, README update, and the test page currently are missing for release state.
Tech review is in progress
If it will be easier for you and review to make these updates separately from the main part of the code, fill free to let me know to create an epic branch.

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  /** Unsubscribes from the observed target {@link Element} changes */
+  public override removeEventListener(callback: EventListener): void;
+  public override removeEventListener(event: SwipeEventName, callback: EventListener): void;
+  public override removeEventListener(event: any, callback: EventListener = event): void {
+    super.removeEventListener(event, callback);
+    if (this.getEventListeners().length > 0) return;
+
+    this.target.removeEventListener('pointerdown', this.handleStart.bind(this), false);
+    this.target.removeEventListener('pointerup', this.handleEnd.bind(this), false);
```
ala-n, line 178:
```suggestion
    this.target.removeEventListener('pointerdown', this.handleStart.bind(this), false);
    this.target.removeEventListener('pointerup', this.handleEnd.bind(this), false);
```
Incorrect the subscription will not happen.
According to the `bind` method spec, it returns a NEW (every time) function that freezes the original function call context to the passed object.
This new function(s) is not subscribed to the element, so it can not be unsubscribed.
Consider @bind decorator, constructor binding, or better use internal ESLEventListener.subscribe API to create internal subscription using the ESL Event Utils module capabilities

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,180 @@
...
+    }
+    if (Math.abs(yDiff) > swipeThreshold && timeDiff < this.config.timeout) {
+      return yDiff > 0 ? 'up' : 'down';
+    }
+
+    return 'left';
+  }
+
+  /** Unsubscribes from the observed target {@link Element} changes */
+  public override removeEventListener(callback: EventListener): void;
```
ala-n, line 171: Cosmetic: place `removeEventListener` and `addEventListener` nearby for readability reasons. When two "symmetrical" methods paced together, it is easier to manage internal actions and see if there is an inconsistency in it

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Saves swipe start event target, time when swipe started, pointerdown event and coordinates.
+   * @param e - pointer event
+   */
+  protected handleStart(e: PointerEvent): void {
+    this.startEl = e.target as HTMLElement;
+    this.timeDown = e.timeStamp;
+    this.startEvent = e;
+    this.xDown = e.clientX;
+    this.yDown = e.clientY;
```
ala-n, line 99: We already have an event, do we need to extract point coordinates and timestamps here?

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   * @param $el - the element to listen for swipe events on
+   * @param threshold - the distance threshold (px, vh, vw supported) that must be traveled before a swipe event is triggered. Optional. Defaults to 20px
+   * @param timeout - the time in milliseconds that must elapse between pointerdown and pointerup events before a wipe event is triggered. Optional.
+   * Defaults to 500.
+   * @returns Returns the instance of ESLSwipeEventTarget {@link ESLSwipeEventTarget}.
+   */
+  public static for($el: Element | ESLMixinElement, threshold?: string, timeout?: number): ESLSwipeEventTarget {
```
ala-n, line 71: In case we face the situation when we need more than two setting attributes here we will need to create a config object anyway (but it will be a Breaking change for users)

Recommend considering the following API for users:
```
  ESLSwipeEventTarget.for($el: ESLDomElementTarget, options: Partial<SwipeEventTargetConfig>): ESLSwipeEventTarget {
```

Please also note that we use the ESLDomElementTarget interface here and `resolveDomTarget(target)` to resolve it to the Dom element.
First of all, ESL as a library uses a much more wide definition in API than just Element / ESLMixinElement
Also, under the hood, it does not reach the ESLMixinElement at all due to all ESL modules should have the most independent from other modules' state to maintain optimal tree shaking for every ESL user.

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    if (this.config.units === 'vh') {
+      swipeThreshold = Math.round((swipeThreshold / 100) * document.documentElement.clientHeight); // get percentage of viewport height in pixels
+    }
+    if (this.config.units === 'vw') {
+      swipeThreshold = Math.round((swipeThreshold / 100) * document.documentElement.clientWidth); // get percentage of viewport width in pixels
+    }
+
+    return swipeThreshold;
+  }
```
ala-n, line 147: Could you consider creating the following API for that method?
```
parseTreshold(value: `${numder}${''|'px'|'vh'|'vw'}`): numder
```

It doesn't look like it will slow down the check if we call it in place. We need to get the exact threshold amount (as a number in px)
In that case, we also fix the incorrect behavior that is right now related to viewport resize (if we parse it in the constructor, we will not reflect any type of resize event)

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,180 @@
...
+  distanceY: number;
+  touchStartOriginalEvent: PointerEvent;
+  touchEndOriginalEvent: PointerEvent;
+}
+
+interface SwipeEventTargetConfig {
+  threshold: number;
+  units: string;
+  timeout: number;
+}
```
ala-n, line 20: Interfaces are usually a part of Public API that the user faces first so we need to make TS/JS Docs for it.

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,180 @@
+import {SyntheticEventTarget} from '../../../esl-utils/dom/events/target';
+import {ESLMixinElement} from '../../../esl-mixin-element/ui/esl-mixin-element';
+import {overrideEvent} from '../../../esl-utils/dom/events/misc';
+
+type SwipeDirection = 'left' | 'right' | 'up' | 'down';
+type SwipeEventName = 'swipe' | 'swipe:left' | 'swipe:right' | 'swipe:up' | 'swipe:down';
+
+interface SwipeEvent {
```
ala-n, line 8: We can make an exact Event class to provide both typings and creation incapsulation.

See `resize.adapter.event.ts`, `resize.adapter.ts` , and `decorated.target.ts` implementations, as they will have a lot of similarities in some implementation moments.

ala-n (changes requested):

src/modules/esl-event-listener/core/targets/swipe.target.ts
grechihinrhp: It's used to calculate distance and time between start and end points and based on config make a decision weather to trigger swipe event or not

ala-n: Yes that's clear, but we have the `pointdown` event stored and available in `pointerup` handler
Event is finalized object, so `e.clientX` will not change in `pointerup` handler.

src/modules/esl-event-listener/core/targets/swipe.target.ts
grechihinrhp: Moved threshold resolving to separate helper function `resolveCSSSize` under `esl-utils/dom/units`. Created tests for new helper.

src/modules/esl-utils/dom/units.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * @param value - CSS value string in px, vh, vw or without units. {@link CSSSize}
+ * @returns number in pixels.
+ */
+export const resolveCSSSize = (value: CSSSize): number => {
+  let num = parseInt(value, 10);
+  const units = value.replace(num.toString(), '');
+
+  if (units === 'vh') {
+    num = Math.round((num / 100) * document.documentElement.clientHeight); // get percentage of viewport height in pixels
```
ala-n, line 12:
```suggestion
    return Math.round((num / 100) * document.documentElement.clientHeight); // get percentage of viewport height in pixels
```

src/modules/esl-utils/dom/units.ts
```diff
@@ -0,0 +1,19 @@
...
+  }
+  if (units === 'vw') {
+    num = Math.round((num / 100) * document.documentElement.clientWidth); // get percentage of viewport width in pixels
```
ala-n, line 15:
```suggestion
    return Math.round((num / 100) * document.documentElement.clientWidth); // get percentage of viewport width in pixels
```

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,144 @@
...
+      };
+
+      // fire `swipe` event on the element that started the swipe
+      this.dispatchEvent(ESLSwipeGestureEvent.fromConfig('swipe', swipeInfo));
+      // fire `swipe:${dir}` event on the element that started the swipe
+      this.dispatchEvent(ESLSwipeGestureEvent.fromConfig(`swipe:${direction}` as SwipeEventName, swipeInfo));
+    }
+
+    // reset values
+    this.startEl = null;
```
ala-n, line 94:
```suggestion
    this.startEvent= null;
```

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Triggers swipe event {@link SwipeEventName} with details {@link ESLSwipeGestureEvent} when pointerup event
+   * occurred and threshold and distance match configuration
+   * @param e - pointer event
+   */
+  @bind
+  protected handleEnd(e: PointerEvent): void {
+    // if the user released on a different target, cancel!
+    if (this.startEl !== e.target) return;
```
ala-n, line 74:
```suggestion
    if (!this.startEvent || this.startEvent.target !== e.target) return;
```

julia-murashko (approved):

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,148 @@
...
+
+import type {CSSSize} from '../../../esl-utils/dom/units';
+import type {SwipeDirection, ESLSwipeGestureEventInfo, SwipeEventName} from './swipe.event';
+import type {ESLDomElementTarget} from '../../../esl-utils/abstract/dom-target';
+
+/**
+ * Describes parsed configuration of {@link ESLSwipeGestureTarget}
+ */
+interface SwipeEventTargetConfig {
+  threshold: string;
```
dshovchko, line 17: Perhaps this should be of the CSSSize type?

src/modules/esl-utils/dom/test/units.test.ts
```diff
@@ -0,0 +1,25 @@
+import {resolveCSSSize} from '../units';
+
+describe('resolveCSSSize tests', () => {
```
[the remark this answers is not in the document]
grechihinrhp: Thanks! Updated tests

src/modules/esl-event-listener/core/targets/swipe.event.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Describes swipe information provided with {@link ESLSwipeGestureEvent}
+ */
+export interface ESLSwipeGestureEventInfo {
+  target: Element;
+  /** Swipe direction {@link SwipeDirection} */
+  direction: SwipeDirection;
+  /** Distance between the points where pointerdown and pointerup events occurred along the x axis */
+  distanceX: number;
+  /** Distance between the points where pointerdown and pointerup events occurred along the x axis */
```
yadamskaya, line 22:
```suggestion
  /** Distance between the points where pointerdown and pointerup events occurred along the y axis */
```

yadamskaya (approved):

NastaLeo (approved):

src/modules/esl-event-listener/core/targets/swipe.target.ts
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Unsubscribes from the observed target {@link Element} changes
+   */
+  public override removeEventListener(callback: EventListener): void;
+  public override removeEventListener(event: SwipeEventName, callback: EventListener): void;
+  public override removeEventListener(event: any, callback: EventListener = event): void {
+    super.removeEventListener(event, callback);
+    if (this.getEventListeners().length > 0) return;
+
+    ESLEventUtils.unsubscribe(this.target, 'pointerdown');
```
NastaLeo, line 164: I guess, here we should specify handler
ESLEventUtils.unsubscribe(this.target, this.handleStart);
ESLEventUtils.unsubscribe(this.target, this.handleEnd);

ala-n (commented):
A little addition to @NastaLeo comment above.
The host itself is not correct, in case we fix the host - we can simplify unsubscription

ala-n (approved):

Issue #1823, opened 2023-07-21
[🚀esl-share]: add out of the box share config default


ala-n:
As an ESL consumer, I want to be able to use out-of-the-box share buttons configuration.

Also, consider merging algorithms to simplify the customization process from the client side.

PR #1846, opened 2023-08-04
[:rocket: esl-share]: add out of the box share config default

dshovchko:
Closes: #1823


src/modules/esl-share/core/esl-share-config-buttons.ts
```diff
@@ -0,0 +1,55 @@
+import type {ESLShareConfigShape, ESLShareButtonConfig, ESLShareGroupConfig} from './esl-share-config';
+
+/** Gets item of the section by name */
+function getConfigSectionItem<T extends ESLShareButtonConfig | ESLShareGroupConfig>(section: T[], name: string): T | undefined {
+  return section.find((item) => item.name === name);
+}
+
+/** Class for managing share buttons configurations */
+export class ESLShareConfigButtons {
```
ala-n, line 9: Can't we just use ESLShareConfig for that purpose?

julia-murashko (approved):

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -32,31 +34,91 @@ export interface ESLShareGroupConfig {
...
+
+  /**
+   * Sets config with a promise of a new config object or using a config provider function.
+   * @returns Promise of the current config
+   */
+  public static set(provider?: ESLShareConfigProviderType | ESLShareConfigShape): Promise<ESLShareConfig> {
+    if (typeof provider === 'function') ESLShareConfig._config = provider().then(ESLShareConfig._from);
+    if (typeof provider === 'object') ESLShareConfig._config = Promise.resolve(ESLShareConfig._from(provider));
+    return ESLShareConfig.get();
+  }
```
ala-n, line 77: It doesn't look like we support the primary use case for consumers:
We definitely can provide default settings for buttons, but we 100% can not provide the groups for the end user
So we need a merging strategy instead

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -1,5 +1,7 @@
-/** {@link ESLShareConfig} provider type definition */
-export type ESLShareConfigProviderType = () => Promise<ESLShareConfig>;
+import defaultConfig from '../config/default.json';
```
ala-n, line 1: It looks like it's a strict dependency on the config module 
So that means a user requiring a short config can not reduce the bundle size...
It is not acceptable according to our acceptance criteria:
The user can use the default provided by ESL if he wants, but he should not if he doesn't want to.

ala-n (changes requested):

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -32,31 +34,91 @@ export interface ESLShareGroupConfig {
 }
 
 /** The definition of `ESLShare` component configuration */
-export interface ESLShareConfig {
+export interface ESLShareConfigShape {
```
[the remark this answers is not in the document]
dshovchko: Done

src/modules/esl-share/core/esl-share-config.ts
dshovchko: This has also been changed. You can now add buttons and groups to the config, or override an existing one.
That is, it is quite possible in runtime, for example, to override the group of buttons depending on the locale.

src/modules/esl-share/buttons/wykop.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const wykop = {
+  action: 'media',
+  icon: '<svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" role="presentation" viewBox="0 0 32 32"><path d="M14.722 19.784l-3.48-6.832-2.667 1.36 3.82 7.497.056-.03.313.61 10.608-5.404-.48-.944-3.65-7.165-2.667 1.36 3.48 6.83-1.332.68-3.48-6.832-2.666 1.358 3.48 6.832-1.332.68z"/><path d="M7.372 12.77c0-2.38 1.86-4.308 4.152-4.308h8.952c2.294 0 4.152 1.928 4.152 4.308v6.46c0 2.38-1.86 4.308-4.152 4.308h-8.952c-2.294 0-4.152-1.928-4.152-4.308v-6.46zM5 12.77v6.46C5 22.97 7.92 26 11.524 26h8.952C24.08 26 27 22.97 27 19.23v-6.46C27 9.03 24.08 6 20.476 6h-8.952C7.92 6 5 9.03 5 12.77z"/></svg>',
+  iconBackground: '#fb803f',
+  link: '//www.wykop.pl/dodaj?url={u}&title={t}',
+  name: 'wykop',
+  title: 'Wykop'
+};
+ESLShareConfig.instance.appendButton(wykop as unknown as ESLShareButtonConfig);
```
ala-n, line 15: Why do we need to drop the initial object type? that doesn't look correct due to use should do the same without any blocks from the type perspective.

src/modules/esl-share/buttons/pusha.ts
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable max-len */
```
ala-n, line 1:
```suggestion
```
What was the reason?

src/modules/esl-share/test/config/reference-buttons.ts
```diff
@@ -0,0 +1,164 @@
+/* eslint-disable max-len */
```
ala-n, line 1:
```suggestion
```

src/modules/esl-share/test/config/reference-buttons.ts
```diff
@@ -0,0 +1,164 @@
...
+export const referenceButtons = {
```
ala-n, line 2: Do we have any benefit from a plain copy of the inner config?
Not sure this gives any additional advantage for checking buttons config. 

Test for import, config size, or consistency with the object inside the config file or general type shate - that's a valid test for sure.
 But if we have a problem with a URL pattern or icon inside the button config, I believe right now we just have it twice: in test data and in the real data :)

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -31,32 +35,137 @@ export interface ESLShareGroupConfig {
...
+  const index = section.findIndex((i) => i.name === item.name);
+  if (index >= 0) section[index] = item;
+  else section.push(item);
+}
+
+/** Class for managing share buttons configurations */
+export class ESLShareConfig extends SyntheticEventTarget {
+  @memoize()
+  public static get instance(): ESLShareConfig {
+    return new ESLShareConfig();
```
ala-n, line 54: Cosmetic: possibly we need to declare constructor private/protected then

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -31,32 +35,137 @@ export interface ESLShareGroupConfig {
...
+  /**
+   * Sets config with a promise of a new config object or using a config provider function.
+   * Each of the buttons and groups specified in the new config will be appended to the current config.
+   * @returns Promise of the current config
+   */
+  public static set(provider?: ESLShareConfigProviderType | ESLShareConfig): Promise<ESLShareConfig> {
+    if (typeof provider === 'function') return provider().then(ESLShareConfig.append);
+    if (typeof provider === 'object') ESLShareConfig.append(provider);
+    return Promise.resolve(ESLShareConfig.instance);
+  }
```
ala-n, line 66: Cosmetic
```suggestion
  public static async set(
    provider?: ESLShareConfigProviderType | Promise<ESLShareConfig> | ESLShareConfig
  ): Promise<ESLShareConfig> {
    if (typeof provider === 'function') return this.set(provider())
    if (typeof provider === 'object') ESLShareConfig.append(provider instanceof Promise ? await provider : provider);
    return ESLShareConfig.instance;
  }
```

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -31,32 +35,137 @@ export interface ESLShareGroupConfig {
...
+
+  /** @returns config of buttons */
+  public get buttons(): ESLShareButtonConfig[] {
+    return this._buttons;
+  }
+
+  /** @returns config of groups */
+  public get groups(): ESLShareGroupConfig[] {
+    return this._groups;
+  }
```
ala-n, line 92:
```suggestion
  public readonly buttons: ESLShareButtonConfig[] = [];
  public readonly groups: ESLShareGroupConfig[] = [];
```

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -31,32 +35,137 @@ export interface ESLShareGroupConfig {
...
+    super.addEventListener('change', callback);
+  }
+
+  public override removeEventListener(callback: EventListener): void;
+  public override removeEventListener(event: 'change', callback: EventListener): void;
+  public override removeEventListener(event: any, callback: EventListener = event): void {
+    if (typeof event === 'string' && event !== 'change') return;
+
+    super.removeEventListener('change', callback);
+  }
```
ala-n, line 170: Potentially, we can use built-in declaration.
Due to it's mostly inner objects that do not pretend to be observed outside the ESLShare module, we can ignore incorrect subscriptions without special warnings.

src/modules/esl-share/core/esl-share.ts
```diff
@@ -161,4 +157,10 @@ export class ESLShare extends ESLBaseElement {
   protected storePopup(value: ESLPopup): void {
     (this.constructor as typeof ESLShare)._popupStore.set(this.list, value);
   }
+
+  @listen({event: 'change', target: ESLShareConfig.instance})
+  @decorate(debounce, 150)
```
ala-n, line 162:
```suggestion
```
Over-debounced, it is rare event by architecture

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -31,32 +35,137 @@ export interface ESLShareGroupConfig {
...
+   * Appends group (inserts or updates) to the current config.
+   * @returns config instance
+   */
+  public appendGroup(group: ESLShareGroupConfig): ESLShareConfig {
+    setConfigSectionItem(this.groups, group);
+    this._onUpdate();
+    return this;
+  }
+
+  @decorate(debounce, 25)
```
ala-n, line 148:
```suggestion
  @decorate(microtask)
```

NastaLeo (approved):

src/modules/esl-share/buttons/wykop.ts
dshovchko: Sorry, I was inattentive and missed that.

src/modules/esl-share/core/esl-share-config.ts
dshovchko: done

src/modules/esl-share/core/esl-share-config.ts
dshovchko: At first, I had some problems with listeners... The overrides solved the problem. But now it turned out to be unnecessary. And I prefer the option without them.

src/modules/esl-share/test/config/reference-buttons.ts
dshovchko: The benefit was when I transferred the configuration from JSON to typescript files. I managed to make 5 mistakes in values on 20 files (don't ask how! :smiling_face_with_tear:  ). These tests helped to identify errors in the values. But, when updating the existing buttons or adding new buttons, they do not make sense, I agree with you on this. Therefore, I deleted them.

src/modules/esl-share/core/esl-share.ts
```diff
@@ -161,4 +157,10 @@ export class ESLShare extends ESLBaseElement {
...
+  protected onConfigChange(): void {
+    this.init();
```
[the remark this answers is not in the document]
dshovchko: Thanks! Fixed

src/modules/esl-share/buttons/pusha.ts
dshovchko: The lines of button configs are extra long. They exceed 160 symbols.

src/modules/esl-share/test/buttons/button-test-plan-factory.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    test(`"${actionName}" action was registered when importing the button`, () => {
+      expect(ESLShareActionRegistry.instance.has(actionName)).toBe(true);
+    });
+  };
+}
+
+export function createButtonMatchingTestPlan(name: string, action: string): () => void {
+  return () => {
```
ala-n, line 17: Cosmetic and can be moved to the separate update:
Jest has built in tools to declare shape validation rules
```ts
interface SomeObjType {
    a: number;
    b: string;
    subObj: {
      c: number;
    };
    tuple: [string, string];
    optional?: string;
  }

  const SomeObjTypeShape: Readonly<SomeObjType> = expect.objectContaining({
    a: expect.any(Number),
    b: expect.any(String),
    subObj: expect.objectContaining({
      c: expect.any(Number),
    }),
    tuple: expect.arrayContaining([expect.any(String), expect.any(String)])
    // Skip optional propertys
  });

  test('Shape match sample', async () => {
    const objLikeShape = {
      a: 1,
      b: '2',
      subObj: {
        c: 3
      },
      tuple: ['4', '5']
    };

    expect(objLikeShape).toEqual(SomeObjTypeShape);
    expect({a: 1}).not.toEqual(SomeObjTypeShape);
  });
```

so all non-dunamic (non parametrized) tests may look like
```ts
   test(
      `The button for ${name} accepts ESLShareButtonConfig shape`, 
      () => expect(ESLShareConfig.instance.getButton(name)).toEqual(ESLShareButtonConfigShape);
   );
```

ala-n (approved):

src/modules/esl-share/test/buttons/button-test-plan-factory.ts
dshovchko: Thanks a lot! I will look at it in deep in the next waves.

Natalie-Smirnova (approved):

PR #1873, opened 2023-08-21
feat(esl-utils): make __esl_sequanse__ a private key using Symbol

HenadzV:
<body>
make __esl_sequanse__ a private key using Symbol

Closes: #1866

### DIFF
--- a/src/modules/esl-utils/misc/uid.ts
+++ b/src/modules/esl-utils/misc/uid.ts
@@ -1,11 +1,11 @@
-const eslSequencesKey = Symbol('__esl_sequences__');
+const SEQUENCE_KEY = Symbol('__esl_sequences__');
 const ns = window || global;
-const sequences = ns[eslSequencesKey] || new Map<string, number>();
-ns[eslSequencesKey] = sequences;
+const sequences = ns[SEQUENCE_KEY] || new Map<string, number>();
+ns[SEQUENCE_KEY] = sequences;
 
 declare global {
   interface Window {
-    [eslSequencesKey]: Map<string, number>;
+    [SEQUENCE_KEY]: Map<string, number>;
   }
 }
 
--- a/src/modules/esl-utils/misc/uid.ts
+++ b/src/modules/esl-utils/misc/uid.ts
@@ -1,4 +1,4 @@
-const SEQUENCE_KEY = Symbol('__esl_sequences__');
+const SEQUENCE_KEY: unique symbol = ((window.Symbol || String)('__esl_sequences__')) as any;
 const ns = window || global;
 const sequences = ns[SEQUENCE_KEY] || new Map<string, number>();
 ns[SEQUENCE_KEY] = sequences;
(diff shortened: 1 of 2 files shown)


src/modules/esl-utils/misc/uid.ts
```diff
@@ -1,10 +1,11 @@
+const eslSequencesKey = Symbol('__esl_sequences__');
```
ala-n, line 1: For 4th version, we still use fallback
```suggestion
const eslSequencesKey = (window.Symbol || String)('__esl_sequences__');
```
Also we can simplify variable name and consider it as a constant e.g. `KEY` or `SEQUENCE_KEY`

ala-n (approved):

src/modules/esl-utils/misc/uid.ts
abarmina: SEQUENCE_KEY - it's a good idea

abarmina (approved):

fshovchko (approved):

dshovchko (approved):

NastaLeo (approved):

ala-n (changes requested):
@HenadzV please resolve the comment to unlock the PR

ala-n:
Note for @exadel-inc/esl-core-maintainers : 
Add the `require squash` marker to modify the commit message to the correct change state ( -> refactor)

src/modules/esl-utils/misc/uid.ts
HenadzV: If we use ```(window.Symbol || String)('__esl_sequences__');``` we get  "TS7015: Element implicitly has an 'any' type because index expression is not of type 'number'" and "TS1169: A computed property name in an interface must refer to an expression whose type is a literal type or a 'unique symbol' type" errors. I am not sure how to handle them quickly and we currently use the same syntax ```/** Key to store listeners on the {@link SyntheticEventTarget} instance*/
const LISTENERS = Symbol('_listeners');``` in target.ts. If it is really important I`ll spend some more time handling errors.

Natalie-Smirnova (approved):

src/modules/esl-utils/misc/uid.ts
```diff
@@ -1,10 +1,11 @@
+const SEQUENCE_KEY = Symbol('__esl_sequences__');
```
ala-n, line 1:
```suggestion
const SEQUENCE_KEY = (window.Symbol || String)('__esl_sequences__');
```
As mentioned above, we can't declare symbols without fallback until 5th version with official IE support drop released

src/modules/esl-utils/misc/uid.ts
ala-n: To resolve type issue declare it explicietly as unique symbol
```
const SEQUENCE_KEY: unique symbol = ((window.Symbol || String)('__esl_sequences__')) as any;
```

yadamskaya (approved):

ala-n (approved):

abarmina (approved):

PR #1906, opened 2023-08-29
refactor(esl-event-listener): internal API and structure cleanup for `esl-event-listener` module

ala-n:


fshovchko (approved):

src/modules/esl-event-listener/core/api.ts
```diff
@@ -41,9 +40,7 @@ export class ESLEventUtils {
    * @param host - host object (listeners context) to associate subscription
```
NastaLeo, line 40: Mb replace this comments to ESLEventListener and for unsubsrive remove as doubled.

NastaLeo (approved):

src/modules/esl-event-listener/core/api.ts
ala-n: Thanks!
Need to check if they automatically inherited in IDE spelling in case yes, we can consider shortcutting descriptions here
Add to the todo list for the follow ups

dshovchko (approved):

abarmina (approved):

PR #1917, opened 2023-09-05
feat(site): improve performance by replacing page content

fshovchko:


ala-n:
Just a small note: 
If we see that the service worker is going to give us a benefit -
for the result version, it should use the existing code environment (TS + webpack). 
You don't need to make a separate folder and use a plain JS single file, and it's still a client-side code.

pages/src/route/route-link.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  public static override register(): void {
+    ESLDemoRouterService.init();
+    super.register();
+  }
+
+  @listen('click')
+  protected async _onClick(e: Event): Promise<void> {
+    e.preventDefault();
+    if (!this.href || this.href === location.pathname) return;
```
dshovchko, line 20: There is a wrong condition here. It works only for non-HTMLAnchorElements, because href attribute on HTMLAnchorElements always equals absolute URL not relative.
```suggestion
    if (!this.href || this.href === location.href) return;
```
or
```suggestion
    if (!this instanceof HTMLAnchorElement || this.pathname === location.pathname) return;
```

pages/src/route/route-link.ts
```diff
@@ -0,0 +1,23 @@
...
+    ESLDemoRouterService.routeContent(this.href);
```
dshovchko, line 21: The signature of routeContenr requires pathname, but you pass the full URL.

pages/src/route/router-service.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  public static get instance(): ESLDemoRouterService {
+    return this.init();
+  }
+
+  public static async routeContent(pathname: string, pushState = true): Promise<void> {
+    const router = ESLDemoRouterService.instance;
+    try {
+      const response = await fetch(pathname, {method: 'GET'});
```
dshovchko, line 24: {method: 'GET'} is default option
```suggestion
      const response = await fetch(pathname);
```

pages/src/route/router-service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export class ESLDemoRouterService {
+  public contentSelector = '#content-routable';
+
+  public previousUrl: URL;
+
+  @memoize()
+  public static init(): ESLDemoRouterService {
+    const router = new ESLDemoRouterService();
+    router.previousUrl = new URL(location.href);
```
dshovchko, line 12: Location is similar to URL object. Maybe it is enough to use a Location object?
or simplify
```suggestion
    router.previousUrl = new URL(location);
```

Issue #1836, opened 2023-07-27
[🐛 site]: open external links in a new window


abarmina:
As a content maintainer, I expect that all external links open in a new window as this is the 'best practice'.

Therefore, links to the internal site (other articles and pages) should open in the same window, but links to external resources should open in a new window.

PR #1918, opened 2023-09-06
feat(docs): open external links in a new window

HenadzV:
<body>
added check for links leading to external resources and add them attribute to open in a new window

Closes:  #1836


ala-n (approved):

NastaLeo (approved):

Natalie-Smirnova (approved):

pages/11ty/markdown.shortcut.js
```diff
@@ -56,11 +56,17 @@ class MDRenderer {
...
-      console.log(color.yellow(`Rewrite link "${link.href}" to "${resultPath}"`));
-      link.href = resultPath;
+    dom.querySelectorAll('a[href]').forEach((link) => {
+      if (link.href.startsWith('.')) {
+        const absolutePath = path.join(path.dirname(basePath), link.href).replace(/\\/g, '/');
+        const resultPath = MDRenderer.processRewriteRules(absolutePath);
+        console.log(color.yellow(`Rewrite link "${link.href}" to "${resultPath}"`));
+        link.href = resultPath;
+      }
+      if (link.href.startsWith('http:') || link.href.startsWith('https:')) {
```
dshovchko, line 66: Can be simplified:
```suggestion
      if (['http:', 'https:'].includes(link.protocol)) {
```

dshovchko (approved):

abarmina (approved):

pages/11ty/markdown.shortcut.js
```diff
@@ -56,11 +56,17 @@ class MDRenderer {
...
+        link.target = '_blank';
+        link.rel = 'noopener norefferer';
```
abarmina, line 68:
```suggestion
        link.rel = 'noopener noreferrer';
```

ala-n: By the way, here is one more important point:
It will be better to remove `noreferrer` at all. We have not many cases with any other domains than Github, especially in the current implementation when we have this processor on the MD transformation level.
If we are talking about GitHub domain we SHOULD NOT block any analytics collecting for the GItHub statistics and any SEO counting for esl-ui.com - github.com relationship.

dshovchko: I guess value `nofollow` is more critical than `noreferrer`. Value `noreferrer` hides information about the site where you come from. I don't see anything criminal to show this information. But `nofollow` value is used when you want to prevent the search engine from transferring the "search ranking" to another page.

ala-n (approved):

yadamskaya (approved):

PR #1958, opened 2023-09-26
feat(esl-event-listener): add `condition` descriptor property to legally prevent subscription

ala-n:
Closes: #1947

### DIFF
--- a/src/modules/esl-event-listener/core/api.ts
+++ b/src/modules/esl-event-listener/core/api.ts
@@ -1,5 +1,7 @@
 import {ExportNs} from '../../esl-utils/environment/export-ns';
+import {resolveProperty} from '../../esl-utils/misc/functions';
 import {dispatchCustomEvent} from '../../esl-utils/dom/events/misc';
+
 import {ESLEventListener} from './listener';
 import {getAutoDescriptors, isEventDescriptor, initDescriptor} from './descriptors';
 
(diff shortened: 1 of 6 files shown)


src/modules/esl-event-listener/core/api.ts
```diff
@@ -9,6 +11,10 @@ import type {
   ESLListenerDescriptorFn
```
NastaLeo, line 11: missed string import * from './types' as I understood

NastaLeo (approved):

src/modules/esl-event-listener/core/api.ts
ala-n: Unfortunately export != import to current file even if we declare explicitly the export members

NastaLeo: Yep) Sorry for being inattentive

julia-murashko (approved):

yadamskaya (approved):

src/modules/esl-event-listener/test/listener.resubscribtion.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    @listen({event: 'click', condition: (that: TestSubsEl) => that.enabled})
+    onClick() {
+      this.mockFn();
+    }
+  }
+
+  TestSubsEl.register();
+
+  const host = TestSubsEl.create();
+  test('ESLEventListener initial subscription is rejected by condition', () => {
```
abarmina, line 28: very-very cosmetic: pls, add a space before ‘test’, otherwise it's hard to see it

abarmina (approved):

Issue #1529, opened 2023-03-09
[🔧eslint]: make sure both `pages` and `eslint` modules correctly integrated into release process


ala-n:
As an ESL maintainer, I want to have an automatic build for `esl-eslint-plugin` package introduced in the bounds of #1374

Options: 
1. Ideal - auto release in case changes related to plagin presented
2. Release by separate semantic-release run (~1)
3. Release by the option of current release Workflow

Additional requirements:
 - update the version of the package.json accordingly

ala-n:
According to the latest repository update, there is a way to handle submodules without extra steps

Important architectural notes:
1) We are using [NPM workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces)
1.1) ! lerna or yarn are not approved !
2) We should keep both `pages` (internal) and `lint` (public) packages with a consistent (equal) version to the root module
2.1) It's fine to upgrade the version of the submodule without actual code change in the module code
(for now we have just support modules, so that an be assumed as correct behaviour that their version declares compatibility but not the version of the module itself )

ala-n:
@fshovchko feel free to remove the `needs assignee` label in case you are going to work on that ticket.
I just made it back due to a significant scope and strategy change from the initial investigation ticket.

ala-n:
Partially solved in https://github.com/exadel-inc/esl/pull/2003
Moving to 4.14.0 scope as non-blocker

ala-n:
Solved there is no additional step required after https://github.com/exadel-inc/esl/pull/2003

PR #1990, opened 2023-10-17
chore(build): update version bumper

fshovchko:
After looking up the solutions for automatic version update, it was decided to enhance our own already made solution
#1529


ala-n:
It looks like we need to revisit the approach. As was discovered in https://github.com/exadel-inc/esl/pull/2003 the version bumping can be handled via existing npm processes, nevertheless we still need a solution for meta information update which `bump.js` can handle

bump.config.yml
```diff
@@ -0,0 +1,5 @@
+files:
+  - filePath: './eslint/package.json'
+    patterns:
+      - '"name"\s*:\s*"\@exadel\/eslint-plugin-esl"[^{}]*"version"\s*:\s*"(\d+\.\d+\.\d+)"'
+      - '"@exadel\/esl\": \"\^(\d+\.\d+\.\d+)\"'
```
ala-n, line 5: There is no script to call bump.js in the bounds of the root package. Moreover, now, it should not be called in the scope of the root package.

The version control is currently handled by npm built-in script flow.
We need to update the flow of version change of the exact package using the version provided by the npm arguments (but not the root package)
Please consider registering a post (version or postversion) hook inside of eslint module to manage related versions.

ala-n (changes requested):

eslint/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "jest",
     "clean": "rimraf dist",
     "build": "npm run clean && tsc --project tsconfig.json",
-    "version": "node ../build/bump.js"
+    "version": "node ../build/bump.js $npm_package_version"
```
ala-n, line 12: You don't need env $npm_package_version here, because
1. it is already `process.env` while mentioning it here produces problems on windows machines (out of cross-env runner)
2. action in version hook is executed after a native npm version process so actual version is also presented inside current package.json

eslint/bump.config.yml
```diff
@@ -0,0 +1,5 @@
+files:
+  - filePath: 'package.json'
+    patterns:
+      - '"name"\s*:\s*"\@exadel\/eslint-plugin-esl"[^{}]*"version"\s*:\s*"(\d+\.\d+\.\d+)"'
+      - '"@exadel\/esl\": \"\^(\d+\.\d+\.\d+)\"'
```
ala-n, line 5: It appears to be incorrect, as I mentioned earlier. 
The native npm version processes a version without our interventions.

ala-n (changes requested):

ala-n:
After a deeper investigation it appears to be that we no longer require any additional actions at all.
The eslint section in `package.json` is redundant.
Test 4.13.0 build of eslint plugin published. There are no additional issues with versioning.

Issue #1591, opened 2023-04-04
[🚀esl-share]: esl-share release version


ala-n:
As an ESL architect, I want to have finalized and unified version of `esl-share` component.

## Goal / Pre-Requirements
 - support for `esl-share` in a popup version
 - unified API for both implementations of `esl-share` in a list view and popup

## Tech Requirements (based on discussion 04/03/23)
- Public API: `esl-share-list` should be renamed to `esl-share`
- The public API remains the same for `ESLShare` (previously `ESLShareList`)
   -- e.g. `ESLShare.config(...)`
- Tech Proposal: move config and builder methods to separate internal service
  -- e.g. `ESLShareBuilder.build(list: string, $host: ESLShare): ESLShareButtons[];`
  -- e.g. `ESLShareBuilder.setGlobalConfig(provider: async () => ESLShareConfig): void;`
- Introduce `mode` or `view ` attribute with two supported values `list`(default) or `popup`
- Implement popup implementation to support the following
```
<esl-share mode="popup" list="..." share-*="">
    <svg viewBox="0 0 48 48">...</svg>
</esl-share-popup>
```
- Make sure the popup implementation use shared popup instance

ala-n:
TODO: @ala-n

PR #1998, opened 2023-10-19
[🚩esl-share]: rework and cleanup (#1591)

ala-n:
NOTE: The beta state of `esl-share` module will not be changed in bounds of 4.13.0 release. FInal release state cleanup (including documentation) is shifted to 4.14.0 milestone


tsconfig.json
```diff
@@ -18,7 +18,8 @@
       "DOM.Iterable"
     ],
     "rootDir": "src",
-    "outDir": "./"
+    "outDir": "./",
+    "resolveJsonModule": true
```
ala-n, line 22: Do we need it? 
If that's tests, we should limit that option to tests only.
Json resolving is not a native future for both ES Modules (browser) and bundlers, so if TS allows and leaves it in output - it is a problem.

src/modules/esl-share/core/esl-share.ts
```diff
@@ -71,94 +62,59 @@ export class ESLShare extends ESLBaseElement {
...
+  private onReady(): void {
+    this.$$attr('ready', true);
+    this.$$fire(this.SHARE_CHANGED_EVENT, {bubbles: false});
   }
 
-  /** Adds popup element to the popup's store. */
-  protected storePopup(value: ESLPopup): void {
-    (this.constructor as typeof ESLShare)._popupStore.set(this.list, value);
+  @listen({event: 'change', target: ESLShareConfig.instance})
+  protected onConfigChange(): void {
```
ala-n, line 114:
```suggestion
  protected _onConfigChange(): void {
```

src/modules/esl-share/core/esl-share.ts
```diff
@@ -20,36 +17,30 @@ export type ESLShareConfigProviderType = () => Promise<ESLShareConfig>;
...
 export class ESLShare extends ESLBaseElement {
   public static override is = 'esl-share';
-  protected static _config: Promise<ESLShareConfig>;
-  protected static _popupStore: Map<string, ESLPopup> = new Map<string, ESLPopup>();
 
   /** Register {@link ESLShare} component and dependent {@link ESLShareButton} */
   public static override register(): void {
     ESLShareButton.register();
+    ESLSharePopup.register();
     ESLShareTrigger.register();
```
ala-n, line 25: It's funny, but the most critical and painful question appears here.
I am trying to remember if we raised that question in the previous architecture discussions, but we need to make a note somewhere if we do...

As for now, we have ESLTogglable <- ESLPopup <- ESLSharePopoup chain, so if a user uses ESLShare, the whole chain is loaded to his bundle. But what if, as a user, I need just a button/buttons list ...
What strategy do we recommend in that case? (unfortunately ESLPopup chain is not so light to skip it, as a small additional payload)

I cannot say that's a blocker. But if we are going to do something with that in the future, it might be a breaking change.
So we need an argumentation regarding current architecture, or we may have another way (good news, I don't think something significantly changed except the esl-share widget itself)

ala-n:
+ we need to review Site styles (not only current bugs, but also the way of coloring the icons)

`svg path {fill: somecolor; }` - completely unacceptable way (that's the site styles, but correct implementation may affect our defaults for icons, so it is critical to fix)

ala-n:
✅  Live Tests passed from my side. 
✅  Approved for release 4.13.0 inclusion.

dshovchko (approved):

NastaLeo (approved):

HenadzV (approved):

Natalie-Smirnova (approved):

fshovchko (approved):

abarmina (approved):

PR #2006, opened 2023-10-24
refactor(esl-share): revisit `ESLShareConfig` API, create tests and documentation

ala-n:


src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -33,120 +34,133 @@ export interface ESLShareGroupConfig {
...
-        if (groupConfig) return res.concat(this.getList(groupConfig.list));
-      } else {
-        const btnConfig = this.getButton(btnName);
-        if (btnConfig) res.push(btnConfig);
-      }
-      return res;
-    }, [] as ESLShareButtonConfig[]);
+  public get(query: string): ESLShareButtonConfig[] {
+    const terms = this.resolve(query);
+    console.log('[ESLShare]: resolve list %s to %o', query, terms);
```
ala-n, line 121:
```suggestion
```

src/modules/esl-share/core/esl-share.ts
```diff
@@ -29,7 +29,7 @@ export class ESLShare extends ESLBaseElement {
    * @returns Promise of the current config
    * @deprecated alias for ESLShareConfig.set(), will be removed soon
    */
-  public static readonly config = ESLShareConfig.set;
+  public static readonly config = ESLShareConfig.append;
```
ala-n, line 32:
```suggestion
  public static readonly config = ESLShareConfig.set;
```

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -33,120 +34,132 @@ export interface ESLShareGroupConfig {
...
+    if (terms.includes('all')) return this.buttons;
+    const list = terms.map((name) => this.getButton(name));
+    return uniq(list.filter(Boolean) as ESLShareButtonConfig[]);
+  }
+
+  /** Clears the configuration */
+  public clear(): ESLShareConfig {
+    this._buttons.clear();
+    this._groups.clear();
+    return this;
```
dshovchko, line 130: Do we need to notify clients about config changes?

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -33,120 +34,132 @@ export interface ESLShareGroupConfig {
...
+    return Array.from(this._groups.entries()).map(([name, list]) => ({name, list}));
+  }
+
+  /** @returns list of all available buttons */
+  public get buttons(): ESLShareButtonConfig[] {
+    return Array.from(this._buttons.values());
+  }
+
+  /** Normalize list string by removing groups and extra whitespaces */
+  public resolve(query: string): string[] {
```
dshovchko, line 102: Should it be the public method?

src/modules/esl-share/core/esl-share-config.ts
```diff
@@ -33,120 +34,132 @@ export interface ESLShareGroupConfig {
...
+    if (query.includes('all')) return Array.from(this._buttons.keys());
+    const groups = new Set<string>(); // Deduplicate groups
+    while (query.includes('group:')) {
+      query = query.replace(/group:(\S+)/gi, (term: string, name: string) => {
+        if (groups.has(name)) return '';
+        groups.add(name);
```
dshovchko, line 108: Should we be engaged in optimization? It seems to me that this is a content issue and for the component - unnecessary responsibilities.

dshovchko (approved):

src/modules/esl-share/README.md
```diff
@@ -2,98 +2,175 @@
...
+import '@exadel/esl/modules/esl-share/buttons/facebook';
+
+// Register all buttons ootb configurations
+import '@exadel/esl/modules/esl-share/buttons/all';
 ```
 
-The difference between these two methods is obtained only in the fact that in the first case, the configuration instance is returned and in the second case promise is returned.
+Note: configuration of the button or group appends to the existing configuration but overrides the existing one if the button or group with the same name already exists.
+
+In case you need just a littlle fix up the existing configuration you can use ootb button config and then override it:
```
NastaLeo, line 107:
```suggestion
In case you need just a little fix up the existing configuration you can use ootb button config and then override it:
```

src/modules/esl-share/README.md
```diff
@@ -2,98 +2,175 @@
...
+  /** List of share buttons configurations */
+  buttons?: ESLShareButtonConfig[];
+  /** List of share buttons groups configurations */
+  groups?: ESLShareGroupConfig[];
+}
 ```
 
-Thus, we added 3 buttons `facebook`, `linkedin`, `copy` and described the `mygroup` group containing these buttons. You can prepare the button configurations yourself, or you can import them from the `/src/modules/esl-share/config/` directory. There we have prepared for you in advance about 20 configurations for various actions to share.
+The `ESLShareConfigInit` object can be passed to the `ESLShareConfig.set` method.
+It is use full when you have group you configuration in a single json file, or generate it in the runtime
```
NastaLeo, line 131:
```suggestion
It is usefull when you have group you configuration in a single json file, or generate it in the runtime 
```

NastaLeo (approved):

src/modules/esl-share/core/esl-share-config.ts
ala-n: Here, we have a terrible case (infinite loop with super small leak - so the page just freezes) in case users do this. 
And unlike the simple project, we do not have the option to say it's content in case we have a crash.

abarmina (approved):

PR #2014, opened 2023-10-25
chore(site): add github reference

### LINKED ISSUE #1857: [🚀 for site]:  add github reference on the documentation pages
As an ESL site consumer, I want to be able to navigate from the site documentation page to the most related github source.

E.g. from https://esl-ui.com/components/esl-animate/  to https://github.com/exadel-inc/esl/tree/main/src/modules/esl-animate

Please consider page-level metainformation.

TODO: github link design tbd

fshovchko:
The configuration example:
```
---
aside:
     source: blob/main/src/modules/esl-panel/core/esl-panel.ts
---
```
Currently looks as follows:
Screenshot from 2023-10-25 18-55-01

Closes #1857


pages/views/_includes/navigation/related-links.njk
```diff
@@ -3,12 +3,20 @@
     <div class="related-card">
       <h2 class="related-card-heading">{{ title }}</h2>
       <ul class="related-card-list">
-        {% set links = functions.findItemsByName(pages, collections[collection]) %}
-        {% for page in links | released %}
+        {% if collection === 'source'%}
```
ala-n, line 6: As for right now, it is not a collection, it is just a link
Propose to separate template ("macro") for github as it is not a list of links it is a separate link block.
It definitely could reuse CSS, but it is not right to mixup related lists and single link

pages/11ty/url.filter.js
```diff
@@ -0,0 +1,5 @@
+const ghurl = (url) => 'https://github.com/exadel-inc/esl/' + url;
```
ala-n, line 1: Use `site.config.js` to get this information from `site.yml` but not hadcode and duplicate it in code

ala-n (changes requested):

pages/views/_includes/navigation/source-link.njk
```diff
@@ -0,0 +1,14 @@
+{% macro ghLink(link, title) %}
+  {% if link %}
+    <div class="related-card">
+      <h2 class="related-card-heading">{{ title }}</h2>
+      <ul class="related-card-list">
+          <li class="related-card-item">
```
ala-n, line 6: It should not be list... We constantly have one item, and it is called "SOURCE" section. As for right now, it will be strange to present a list.

pages/11ty/url.filter.js
```diff
@@ -0,0 +1,7 @@
+const {github} = require('./site.config');
+
+const ghurl = (url) => github.srcUrl + url;
+
+module.exports = (config) => {
+  config.addFilter('ghurl', ghurl);
+};
```
ala-n, line 7: Please remove the filter if it's only used for string concatenation. We can handle that in the .njk file. Otherwise, ensure the use of appropriate tools to concatenate URLs internally, and make certain that any input URL is acceptable.

dshovchko (approved):

ala-n:
✅ Live Tests Passed

yadamskaya (approved):

PR #2031, opened 2023-11-06
feat(esl-event-listener): inert wheel detector

### LINKED ISSUE #2017: [🚀esl-event-listener]: request for InertWheel event detector
As a consumer, I want to be able to detect "inert" (long wheel) scrolls through the wheel control of the mouse (or alternative, e.g. long swipe on the touchpad).

During the consumer code research, the following snippet was found:
```ts

  @listen('wheel')
  protected _onWheel(event: WheelEvent): void {
    this._onWheelAggregated(event);
  }

  protected _onWheelAggregated = aggregate((events: WheelEvent[]) => {
    const dist = events.reduce((agg, e) => agg + e.deltaY, 0);
    if (dist > WHEEL_TOLERANCE) // Long wheel detected
  }, 100);

```

Proposed to simplify user flow in one of the following ways:

Base (event name, target and options are discussable)
```ts
 @listen({
    event: 'longwheel', 
    target: ($el: Element) => ESLWheelTarget.for($el, {timeout: 100, distance: 400})
  })
  protected _onWheel(event: WheelEvent): void {
     // Long wheel detected
  }
```

Open questions: 
  - Units for distance ... Not sure it is unified in browsers or even in specs.

fshovchko:
In the scope of this PR was created a script, that dispatches `longwheel` when long scrolls are detected
Closes #2017


ala-n:
We need Unit Tests, Doc, and Example to approve it for the main branch.
Let's create an epic branch to split work and unlock the current PR for review.

ala-n:
Marked as `require squash` to make 'feat' commit message more detailed
FYI @exadel-inc/esl-core-maintainers

src/modules/esl-event-listener/core/targets/inert.target.event.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export interface ESLWheelEventInfo {
+  /** Distance scrolled along the x-axis */
+  deltaX: number;
+  /** Distance scrolled along the y-axis */
+  deltaY: number;
+  /** First wheel event */
+  startEvent: WheelEvent;
+  /** Last wheel event */
+  endEvent: WheelEvent;
```
ala-n, line 19: I don't think we interested in original events at least in first and last event specifically.
We can consider a list of all aggregated events, or skip this option for now
CC: @abarmina to make API notes in story ticket

src/modules/esl-event-listener/core/targets/inert.target.event.ts
```diff
@@ -0,0 +1,43 @@
...
+export class ESLWheelEvent extends UIEvent implements ESLWheelEventInfo {
+  public override readonly target: Element;
+
+  public readonly deltaX: number;
+  public readonly deltaY: number;
+  public readonly endEvent: WheelEvent;
+  public readonly startEvent: WheelEvent;
+
+  protected constructor(eventName: WheelEventName, target: Element, wheelInfo: ESLWheelEventInfo) {
+    super(eventName, {bubbles: true, cancelable: true});
```
ala-n, line 34: 1. For a customly-detected sequence of events, the meaning of 'cancel' seems unclear. As no valid definition exists, the event should not be considered cancelable.

2. This custom event is dispatched on a SyntheticEventTarget instance, where there is no definition or capability for such an event to propagate or bubble.

src/modules/esl-event-listener/core/targets/inert.target.ts
```diff
@@ -0,0 +1,101 @@
...
+  /** The maximum duration of the wheel events to consider it inertial */
+  timeout?: number;
+}
+
+/**
+ * Implementation of EventTarget to observe wheel events for inertial scrolling
+ */
+export class ESLWheelTarget extends SyntheticEventTarget {
+  protected static defaultConfig: Required<ESLWheelSetting> = {
+    threshold: '400px',
```
ala-n, line 32: Firsts of all, I don't think we are interested in size units here
And the most important thing, I don't think we handle delta properties of the WheelEvent correct
https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaY

fshovchko: Do we also need to track z-axis delta?

src/modules/esl-event-listener/core/targets/inert.target.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  protected resolveEventDetails(events: WheelEvent[]): ESLWheelEventInfo {
+    const delta = events.reduce((agg, e) => ({
+      x: agg.x + this.calculateScrollPixels(e, 'x'),
+      y: agg.y + this.calculateScrollPixels(e, 'y')
+    }), {x: 0, y: 0});
+
+    const startEvent = events[0];
+    const endEvent = events[events.length - 1];
+    return {startEvent, endEvent, deltaX: delta.x, deltaY: delta.y};
```
ala-n, line 75: Propose to keep all events and let the user decide which one is needed. Or go with minimalistic case for now 
(keep just exact time between first and last event and that's it)

src/modules/esl-event-listener/core/targets/inert.target.event.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Event names that could be triggered by {@link ESLWheelTarget}
+ */
+export type WheelEventName = 'longwheel';
+
+/**
+ * Describes scroll information provided with {@link ESLWheelTarget}
+ */
+export interface ESLWheelEventInfo {
```
ala-n, line 11: Add an axis property to have information which axis exactly produces the event
Dispatch separate events for x and y axises

ala-n (approved):
Approved for the epic branch
Implementation looks really nice 👍🏻

src/modules/esl-event-listener/core/targets/inert.target.event.ts
```diff
@@ -0,0 +1,46 @@
+import {overrideEvent} from '../../../esl-utils/dom/events/misc';
+
+/**
+ * Event names that could be triggered by {@link ESLWheelTarget}
```
dshovchko, line 4:
```suggestion
 * Event name that could be triggered by {@link ESLWheelTarget}
```

dshovchko (approved):

Issue #1856, opened 2023-08-08
[🚀 for site]: display draft item in sidebar navigation in prod mode when it is current page


ala-n:
As an ESL site content editor, I want the current page to be always presented in the site navigation.

Draft pages are entirely removed from navigation and index.
At the same time, all the pages are available on prod, and we can have a reason to reference them for A/B testing, demos, and other temporary purposes.

Proposed to render the currently opened draft page in the page navigation (on the same palaces where the page is presented in DEV mode)

NOTE 1: Please make sure you are not going to hack each place where draft pages are excluded everything should be solved using global filters

NOTE 2: We have no prepared prod watch and prod servers commands in the sections of the script. Make sure you have extended it flexibly and compact

PR #2033, opened 2023-11-08
Display draft item in sidebar navigation in prod mode when it is current page

Natalie-Smirnova:
Closes: #1856


pages/11ty/collections.filter.js
```diff
@@ -0,0 +1,22 @@
+/** get value from object by path */
+const get = (item, path) => path.split('.').reduce((acc, key) => acc && acc[key], item);
+
+/** filter to include only items that have a property with passed value */
+function filter(collection, prop, value) {
+  const matcher = arguments.length === 2 ?
```
dshovchko, line 6: I'm not sure about the proper understanding of njk, but it seems strange to me.
Do we really have to check the number of arguments passed? And what if there is only one?

ala-n: That's the way to strictly separate two usecases of the filter:
1) We do not pass the value (third argument in definition, second in filter usage in njk) 
  - `collection | filter('visible')` -  - so we are searching for items with truthy 'visible' field
  - Same with `collection | exclude('visible')` - we are searching for items with falsy 'visible' field
2) We pass the value to the filter:
 - `collection | filter('visible', true)` -  we are searching for items with exact `true` boolean value
 - `collection | exclude('order', 0)` - we are searching for items with order field not equal to 0 (so false, NaN, etc - are acceptable)

julia-murashko (approved):

dshovchko (approved):

pages/11ty/collections.filter.js
ala-n: Update the logic to look more in a ES6 style, clarify documentation

NastaLeo (approved):

PR #2040, opened 2023-11-08
chore(site): create `lorem-ipsum` custom dummy text element

ala-n:
Starting to move staff from `epic/redesign` branch to stable...


pages/src/text/lorem-ipsum.ts
```diff
@@ -0,0 +1,63 @@
+import {ESLBaseElement, attr} from '@exadel/esl/modules/esl-base-element/core';
```
ala-n, line 1:
```suggestion
import {ESLBaseElement} from '@exadel/esl/modules/esl-base-element/core';
import {attr} from '@exadel/esl/modules/esl-utils/decorators';
```

julia-murashko (approved):

yadamskaya (approved):

ala-n:
@exadel-inc/esl-all-contributors team, there is one common question remaining for that PR:

As you can see, the component is generalized enough. While it may not be a production-valuable component, it could still be utilized for demo content in the development process, not only on the ESL site but also on the customer side.

So, do you think it is a good idea to move the component from the site package to ESL modules?
  - 👍🏻 - yes
  - 👎🏻 - no

The default custom tag name has been set to `esl-lorem-ipsum`.
  - 👀 - that's fine
  - 😕 - we need to consider another naming (possibly `esl-dummy-text` in that case)

Feel free to express your opinion via text in case a quick vote with emojis does not clearly convey your thoughts.

pages/src/text/lorem-ipsum.ts
```diff
@@ -0,0 +1,91 @@
...
+    super.connectedCallback();
+    this.rerender();
+  }
+  protected override disconnectedCallback(): void {
+    this.innerHTML = '';
+    this.style.display = '';
+    super.disconnectedCallback();
+  }
+  protected override attributeChangedCallback(name: string, oldValue: string, newValue: string): void {
+    super.attributeChangedCallback(name, oldValue, newValue);
```
dshovchko, line 42: Does it need to call the empty method?

dshovchko (approved):

pages/src/text/lorem-ipsum.ts
```diff
@@ -0,0 +1,91 @@
+import {ESLBaseElement} from '@exadel/esl/modules/esl-base-element/core';
+import {attr} from '@exadel/esl/modules/esl-utils/decorators';
+
+const WORDS = [
```
dshovchko, line 4: I guess "dictionary" is a more appropriate name.

And it seems to me that if users could import the dictionaries they need and use them, then this would be a very cool feature. Moreover, they could make their own dictionaries and use them.

ala-n: Great idea! I'll take a look how to make this setting available

ala-n:
As a result of discussion final tag name: `esl-random-text`

ala-n:
Closed as moved to https://github.com/exadel-inc/esl/pull/2055

Issue #2012, opened 2023-10-24
[🚀esl-share]: esl-share final follow up points


ala-n:
- [x] Resolve questions with deduplication behavior
 - [x] Visuall issue with displaying a huge list of share buttons in a popup in a small amount od space (e.g. mobile devices)
 - [ ] Revisit minimize code if needed
 - [x] Ready state attribute (for exact inner elements but not component wrapper) and ready state event
 - [x] Observed attributes - must be declared (list property definetly should be observable)
 - [x] (the most problematic question): `esl-share` component looks a little bit overkilling (it just complicates things) 
 - [x] add ExportNS using
 - [x] support types for correct using in Typesript 
 - [ ] unclear style loading (popup)

PR #2048, opened 2023-11-13
[ESL Share]: follow up (wave 1)

dshovchko:
Closes: #2012

### DIFF
--- a/src/modules/esl-share/core.ts
+++ b/src/modules/esl-share/core.ts
@@ -1,9 +1,3 @@
-export type {ESLShareButtonTagShape} from './core/esl-share-button.shape';
-export type {ESLShareListTagShape} from './core/esl-share-list.shape';
-export type {ESLSharePopupTagShape} from './core/esl-share-popup.shape';
-export type {ESLSharePopupTriggerTagShape} from './core/esl-share-popup-trigger.shape';
-export type {ESLShareTagShape} from './core/esl-share.shape';
-
 export * from './core/esl-share';
 export * from './core/esl-share-action';
 export * from './core/esl-share-action-registry';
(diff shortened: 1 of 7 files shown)


src/modules/esl-share/core.ts
```diff
@@ -1,3 +1,9 @@
+export type {ESLShareButtonTagShape} from './core/esl-share-button.shape';
+export type {ESLShareListTagShape} from './core/esl-share-list.shape';
+export type {ESLSharePopupTagShape} from './core/esl-share-popup.shape';
+export type {ESLSharePopupTriggerTagShape} from './core/esl-share-popup-trigger.shape';
+export type {ESLShareTagShape} from './core/esl-share.shape';
```
ala-n, line 5: It might be better to do that inside of exact modules.
The types are not a problem of bundle, and we expect the "good" user will use shortcuts and tree shaking. But it will be more convenient for us to link related things.
⚠️ Discussable

src/modules/esl-share/core/esl-share-list.ts
```diff
@@ -12,8 +13,10 @@ import type {ESLShareButtonConfig} from './esl-share-config';
  *
  * ESLShareList is a custom element to dynamically draw {@link ESLShareButton}s using simplified shared config
  */
+@ExportNs('ShareList')
 export class ESLShareList extends ESLBaseElement {
   public static override is = 'esl-share-list';
+  public static observedAttributes = ['list'];
```
ala-n, line 19:
```suggestion
  public static override observedAttributes = ['list'];
```

dshovchko: This member cannot have an `override` modifier because it is not declared in the base class `ESLBaseElement`.

PR #2052, opened 2023-11-15
ESL Avatar

dshovchko:
Created a new component to display users' avatar


src/modules/esl-avatar/core/esl-avatar.less
```diff
@@ -0,0 +1,35 @@
...
+
+  .esl-avatar-img {
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+  }
+
+  .esl-avatar-text {
+    text-transform: uppercase;
+    line-height: var(--esl-avatar-text-size, 1em);
```
ala-n, line 31: That means 
```suggestion
    line-height: 1em;
```

src/modules/esl-avatar/core/esl-avatar.less
```diff
@@ -0,0 +1,35 @@
...
+    color: var(--esl-avatar-text-color);
+    font-size: var(--esl-avatar-text-size);
```
ala-n, line 33: Propose to move declarations on esl-avatar tag level
```suggestion
    color: inherit;
    font-size: inherit;
```

src/modules/esl-avatar/core/esl-avatar.less
```diff
@@ -0,0 +1,35 @@
+esl-avatar {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  transition: opacity 0.3s ease-in-out;
+  opacity: 0;
+  border-radius: 50%;
+  background: var(--esl-avatar-background);
```
ala-n, line 8: Hm do we need to introduce custom variable for existing CSS property that used once (or can be inherited)...
PS: does not blocks current PR just a point to discuss for final epic

src/modules/esl-avatar/core/esl-avatar.ts
```diff
@@ -0,0 +1,109 @@
...
+@ExportNs('Avatar')
+export class ESLAvatar extends ESLBaseElement {
+  public static override is = 'esl-avatar';
+  public static observedAttributes = ['image-src', 'username'];
+
+  /** Event to dispatch on change of {@link ESLAvatar} */
+  @prop('esl:avatar:changed') public AVATAR_CHANGED_EVENT: string;
+
+  /** Source path of the avatar image */
+  @attr() public imageSrc: string;
```
ala-n, line 14: can be just src potentially in case it uses the same syntax as image src

src/modules/esl-avatar/core/esl-avatar.ts
```diff
@@ -0,0 +1,109 @@
...
+  /** The limit number of letters to be displayed in text-only mode */
+  @attr({defaultValue: 2}) public limit: number;
```
ala-n, line 16: To Discuss (minor): the name of the attribute is not clear enough
Possibly: `max-char`, `truncate`, `abbr-length`, just `length`

src/modules/esl-avatar/core/esl-avatar.ts
```diff
@@ -0,0 +1,109 @@
...
+  /** Policy of loading image that is outside of the viewport */
+  @attr({defaultValue: 'lazy'}) public loading: 'eager' | 'lazy';
+  /** The name of the user for whom the avatar is displayed */
+  @attr({defaultValue: ''}) public username: string;
+
+  /** @readonly Marker of displaying mode with image */
+  @boolAttr({readonly: true}) public withImg: boolean;
+  /** @readonly Marker of displaying mode without image, only text */
+  @boolAttr({readonly: true}) public textOnly: boolean;
```
ala-n, line 25: It looks like that's the same or 'concurrent' property:
so we can use one boolen marker attribute 
or string (enumed) attribute in case we have more then 2 states (e.g. empty)

src/modules/esl-avatar/core/esl-avatar.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  protected override attributeChangedCallback(attrName: string, oldVal: string, newVal: string): void {
+    if (!this.connected || oldVal === newVal) return;
+    this.init(true);
+  }
+
+  /** Initializes the avatar */
+  public init(force?: boolean): void {
+    if (this.ready && !force) return;
```
ala-n, line 50: Do we have a real case for mutex here ?
It looks like it is even harmful:
```
const $avatar = document.querySelector('esl-avatar');
const $copy = $avatar.cloneNode();
$copy.username = 'Copy User';
$something.append($copy); // Oups, it's not a  'Copy User' and it's event not initialized as user expect to use non-deep copy
```

src/modules/esl-avatar/core/esl-avatar.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  /** Actions on image loading error */
+  @bind
+  protected _onImageError(): void {
+    this.buildTextContent();
+  }
+
+  /** Actions on complete init and ready component */
+  private onReady(): void {
+    this.$$attr('ready', true);
```
ala-n, line 98: Ready = initialized or content (image) is loaded ?

ala-n:
Mark as require squash 
Possibly we need reword `[initial implementation of avatar component](https://github.com/exadel-inc/esl/pull/2052/commits/41ca919bcf08749054bb509c87eb2201378dac94)` as the user does not expect to see initial (it is expected to see ready result, released or beta)

julia-murashko (approved):

HenadzV (approved):

src/modules/esl-avatar/core/esl-avatar.less
dshovchko: removed

src/modules/esl-avatar/core/esl-avatar.ts
dshovchko: reworked

PR #2121, opened 2023-12-15
feat(esl-toggleable): introducing a base trigger class in purpose to simplify and make Trigger-Toggleable constructions more API strict in ESL

ala-n:


src/modules/esl-trigger/core/esl-base-trigger.ts
```diff
@@ -0,0 +1,232 @@
...
+  @attr({defaultValue: null}) public a11yLabelInactive: string | null;
+
+  /** Show delay value */
+  @attr({defaultValue: 'none'}) public showDelay: string;
+  /** Hide delay value */
+  @attr({defaultValue: 'none'}) public hideDelay: string;
+
+  /**
+   * Show delay value override for hover.
+   * Note: the value should be numeric in order to delay hover action triggers for correct handling on mobile browsers.
```
NastaLeo, line 46: Propose to rewrite definition as not easy to read
for hoverHideDelay

NastaLeo (approved):

src/modules/esl-trigger/core/esl-trigger.ts
```diff
@@ -1,84 +1,36 @@
...
   /** Target Toggleable {@link ESLTraversingQuery} selector */
   @attr({defaultValue: ''}) public target: string;
   /** Action to pass to the Toggleable. Supports `show`, `hide` and `toggle` values. `toggle` by default */
-  @attr({defaultValue: 'toggle'}) public mode: string;
-
-  /** Click event tracking media query. Default: `all` */
-  @attr({defaultValue: 'all'}) public trackClick: string;
-  /** Hover event tracking media query. Default: `none` */
-  @attr({defaultValue: 'not all'}) public trackHover: string;
+  @attr({defaultValue: 'toggle'}) public override mode: 'toggle' | 'show' | 'hide';
```
abarmina, line 26: but as I see mode is a `@prop` on ESLBaseTrigger class

abarmina (approved):

src/modules/esl-trigger/core/esl-trigger.ts
ala-n: That's right: 
 - we define the `mode` as `toggle` va `@prop` on a base level because
    - we do not want to make publically modifiable property (e.g. attribute) as we have instances of trigger that have no user cases to change the mode
    - we can not move the whole mode logic outside of the base level as it brings a lot of complications in implementing base trigger logic and produces a big breaking change in API
  - we do override the `mode` via `@attr` decorated property as it public API of the ESLTrigger (not the base instance for trigger-like instances starting from that moment)

Issue #1462, opened 2023-02-03
[🔨 release 5.0.0]: Drop legacy aliases of ESL modules from 3-4 versions


ala-n:
As an ESL architect, I don't want to have legacy aliases to be presented in the libraries' new versions.
As the migration tools will be provided from the 4th version, it is requested to drop some legacy aliases in 5th version.

The list of aliases to remove :
- `TraversingQuery`
- `EventUtils`
- `generateUID`
- `deepCompare`
- `ToggleableActionParams`
- `fallback-duration` attribute in shape of `ESLPanel` and `ESLPanelGroup`
- `ESLEventUtils.descriptors`
- `addListener` / `removeListener` aliases of `SynteticEventTarget`
- `TraversingUtils`
- `RTLUtils` (#1499) 

NOTE: te list should be synchronized with the #1283

**Target branch**: `main-beta`

Issue #1461, opened 2023-02-03
[🚩release 5.0.0]: Drop IE11 / ES5 support


ala-n:
As an ESL architect, I don't want to spend team resources to support and not popular and completely legacy targets. 
It is requested to drop IE11 browser support and support of ES5 target in general.

Scope of proposed updates:
 - [x] Get rid of the `full` version of the polyfill for esl-ui.com site renderer
 - [x] Remove shims and targets =ES5 (if such presented) in the site codebase
 - [x] Remove ES5-related OOTB ESL polyfills except `es5-target-shim.ts`

**Target branch**: `main-beta`

PR #2129, opened 2023-12-20
epic: drop IE11 support and legacy properties (`main-next` -> `main-beta`)

ala-n:
Closes: #1462
Closes: #1461


src/modules/esl-utils/misc/array.ts
```diff
@@ -36,14 +38,8 @@ export function unwrap(value: any): any {
...
-export const uniq = <T> (arr: T[]): T[] => {
-  if (arr.length < 2) return arr.slice(0);
-  const result: T[] = [];
-  const set = new Set<T>();
-  arr.forEach((item) => set.add(item));
-  set.forEach((item) => result.push(item));
-  return result;
-};
+export const uniq = <T> (arr: T[]): T[] =>
+  arr.length > 1 ? [...new Set<T>(arr)] : arr.slice(0);
```
NastaLeo, line 42: a'm not sure that it is rigth

NastaLeo (approved):

src/modules/esl-utils/misc/array.ts
dshovchko: It looks really right!
The set will be created from an array and will not contain duplicate entries. Then spread on iteration over entries will create an array.

dshovchko (approved):

src/modules/esl-utils/misc/array.ts
dshovchko: It definitely looks right! 
The set will be created from an array and will not contain duplicate entries. Then spread on iteration over entries will create an array.

NastaLeo: Yes, I was confused by condition, but now i see that it's okay) 
Thanks!

src/modules/esl-media/core/esl-media.shape.ts
```diff
@@ -57,12 +52,10 @@ export type ESLMediaTagShape = ({
 
   /** Define condition {@link ESLMediaQuery} to allow load of media resource */
   'load-condition'?: string;
-  /** Define ESL Traversing Query to find target for load-cls-accepted / load-cls-declined */
-  'load-cls-target'?: string;
-  /** Define class to mark player accepted state */
-  'load-cls-accepted'?: string;
-  /** Define class to mark player declined state */
-  'load-cls-declined'?: string;
+  /** Define class / classes to add when load media is accepted. Supports multiple and inverted classes */
```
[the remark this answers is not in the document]
ala-n: It needs to be updated for the whole .shape file.
Will do that separately if no concerns

PR #2131, opened 2023-12-21
refactor(eslint): restructure eslint module rules, simplify usage

ala-n:


dshovchko (approved):

eslint/src/index.ts
```diff
@@ -43,7 +24,12 @@ export const configs = {
   },
   'default-5': {
     rules: {
-      ...buildDefault(DEPRECATED_4_RULES, 'error')
+      ...buildDefault(DEPRECATED_5_RULES, 'warn')
```
abarmina, line 27: here and below is ‚5’, but what about ‚4’?

abarmina (approved):

NastaLeo (approved):

eslint/src/index.ts
dshovchko: We are going to release 5.0.0 so 4 leave in past

abarmina: oh, it’s ‚default-5’, now I see it

Issue #2161, opened 2024-01-10
[🚀 esl-carousel]: add `pre-active` attribute on the base level API


ala-n:
As an ESL Carousel consumer, I want to be able to access from CSS (e.g., for animation purposes) the slide that is going to be active (at the beginning of the animation).

The attribute name `pre-active` is discussable.

Behavior spec:
1. An attribute appears at the beginning of the slide change animation
2. An attribute available only in the process of animation (should be unset at the end)
3. An attribute should appear only on inactive slides (in multi-slide mode) 

Priority: High
Base Branch: `epic/esl-carousel`

PR #2170, opened 2024-01-15
feat: esl carousel add pre-active marker

julia-murashko:
Closes: #2161

### DIFF
--- a/src/modules/esl-carousel/core/esl-carousel.renderer.ts
+++ b/src/modules/esl-carousel/core/esl-carousel.renderer.ts
@@ -95,6 +95,8 @@ export abstract class ESLCarouselRenderer implements ESLCarouselConfig {
       related: index
     }))) return;
 
+    this.setPreActive(index);
+
     try {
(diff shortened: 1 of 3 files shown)


src/modules/esl-carousel/core/esl-carousel.renderer.ts
```diff
@@ -127,13 +130,25 @@ export abstract class ESLCarouselRenderer implements ESLCarouselConfig {
...
       this.$carousel.slideAt(from + i).active = true;
     }
   }
 
+  public setPreActive(from: number): void {
+    this.clearPreActive();
+    const count = Math.min(this.count, this.size);
+    for (let i = 0; i < count; i++) {
+      this.$carousel.slideAt(from + i).preActive = true;
+    }
```
dshovchko, line 145:
```suggestion
    for (let i = from; i < from + count; i++) {
      this.$carousel.slideAt(i).preActive = true;
    }
```
I guess that would be clearer at a glance.

PR #2183, opened 2024-01-18
ESLPopup, ESLShare: docs updating

dshovchko:


src/modules/esl-popup/README.md
```diff
@@ -9,24 +9,94 @@ Authors: *Dmytro Shovchko*, *Alexey Stsefanovich (ala'n)*.
...
+
+If the size of the popup is inconvenient for correct placement along the minor axis (the popup arrow has large margins, the popup has large margins, or the popup itself is very large), the following will happen. First, the popup will try to position itself as specified by the `offset-arrow` attribute. After that, it checks its starting side (left or top) to see if it is out of bounds. If it does, then the position is adjusted towards the end side by the amount of the overrun. At the same time, the position of the popup arrow is checked. If it goes beyond the permitted limits (popup boundaries + arrow indents), the position adjustment is canceled. After that, the same check is performed for the end side (right or bottom). Does it need to make adjustments? If it is necessary, will this adjustment break the position of the popup arrow? If the adjustment would break the popup arrow, the adjustment is canceled. The popup will be displayed as it is.
```
fshovchko, line 70:
```suggestion
If the size of the popup is inconvenient for correct placement along the minor axis (the popup arrow, or the popup itself, or both have large margins.), the following will happen. First, the popup will try to position itself as specified by the `offset-arrow` attribute. After that, it checks its starting side (left or top) to see if it is out of bounds. If it does, then the position is adjusted towards the end side by the amount of the overrun. At the same time, the position of the popup arrow is checked. If it goes beyond the permitted limits (popup boundaries + arrow indents), the position adjustment is canceled. After that, the same check is performed for the end side (right or bottom). Does it need to make adjustments? If it is necessary, will this adjustment break the position of the popup arrow? If the adjustment would break the popup arrow, the adjustment is canceled. The popup will be displayed as it is.
```

NastaLeo (approved):

src/modules/esl-share/README.md
```diff
@@ -286,13 +286,15 @@ The principle of cascading is similar to CSS variables. The value is searched fr
...
  - `set` - static method that updates the configuration with promise resolved to `ESLShareConfigInit` or promise provider function
+ - `update` - static method that updates items configuration from the list with the specified partial config
  - `buttons` - getter that returns list of all available buttons
  - `groups` - getter that returns list of all available groups
  - `get` - selects the buttons for the given list and returns their configuration
  - `clear` - clears the configuration
  - `getButton` - gets the button configuration for the given name
  - `getGroup` - gets the group of buttons configuration for the given name
  - `append` - updates the configuration with a `ESLShareButtonConfig`(s) or `ESLShareGroupConfig`(s)
+ - `update` - updates items configuration from the list with the specified partial config
```
fshovchko, line 297: Would it make sense to split list into static and instance sub-lists?

dshovchko: It seems to me that it should be consistent for all docs but we have never done it this way.
Proposal to discuss this point.

src/modules/esl-popup/README.md
```diff
@@ -9,24 +9,94 @@ Authors: *Dmytro Shovchko*, *Alexey Stsefanovich (ala'n)*.
...
 
-### ESLPopup Attributes | Properties:
+### ESLPopup
 
-- `position` (string) - popup position relative to the trigger (currently supported: 'top', 'bottom', 'left', 'right' ) ('top' by default)
+What is a popup window? It is a rectangular area with content that is displayed above the main content of the page right next to the element that activated the popup. In other words, it's just a rectangle whose position relative to the trigger element is calculated according to some rules. To use a popup element with ease, you need to know these rules, so here is a brief description of them.
+
+#### Popups arrow
+
+A popup is an element that is displayed quite often along with an arrow that is focused on the trigger element that activated this popup. An arrow is a regular span element inside a popup that has a special class (see `ESLPopup Attributes` section for how to define a custom arrow class) on it. The `ESLPopup` element always checks for the presence of an arrow inside and creates it in case of its absence. So if you need a popup without an arrow, add the `disable-arrow` class to `ESLPopup` element.
```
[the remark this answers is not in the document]
dshovchko: I disagree. The meaning of the sentence is wrong. Perhaps it would be better to rewrite it like this, do you agree?
```suggestion
A popup is an element that is displayed quite often, along with an arrow that focuses on the element that triggered the popup. An arrow is a regular span element inside a popup that has a special class (see `ESLPopup Attributes` section for how to define a custom arrow class) on it. The `ESLPopup` element always checks for the presence of an arrow inside and creates it in case of its absence. So if you need a popup without an arrow, add the `disable-arrow` class to `ESLPopup` element.
```

julia-murashko (approved):

src/modules/esl-share/README.md
dshovchko: done

fshovchko (approved):

src/modules/esl-popup/README.md
```diff
@@ -9,24 +9,94 @@ Authors: *Dmytro Shovchko*, *Alexey Stsefanovich (ala'n)*.
...
+ - instructions for fine-tuning the behavioral positioning;
+ - whether the popup should observe the intersection of the viewport for the entire time it is activated.
+
+##### Optimistic scenario
+
+If there is no space where you set it, then let's assume that you just guessed wrong and all the free space is on the other side or somewhere nearby. So let's let the popup find the optimal location on its own.
+
+Here it is important to choose the behavioral strategy that the popup should follow. It is specified by the `behavior` attribute. Possible values: `fit`, `fit-major`, `fit-minor`, `none`.
+
+The `fit-major` strategy instructs the popup to always try to stay within the viewport but to adjust its positioning only along the major coordinate axis. In practice, it looks like this: if you specify which side of the trigger the popup should be on, and there is no room for it, the popup seems to flip relative to the element trigger. If the position was top, it flips and becomes bottom, if it was left, it becomes right, and vice versa.
```
abarmina, line 54:
```suggestion
The `fit-major` strategy instructs the popup to always try to stay within the viewport but to adjust its positioning only along the major coordinate axis. In practice, it looks like this: if you specify which side of the trigger the popup should be on, and there is no place for it, the popup seems to flip relative to the element trigger. If the position was top, it flips and becomes bottom, if it was left, it becomes right, and vice versa.
```

abarmina (approved):

ala-n (approved):

PR #2184, opened 2024-01-18
feat(attr): add provider to default value in attr

nattallius:


src/modules/esl-utils/decorators/attr.ts
```diff
@@ -17,7 +18,7 @@ type AttrDescriptor<T = string> = {
   /** Use data-* attribute */
   dataAttr?: boolean;
   /** Default property value. Used if no attribute is present on the element. Empty string by default. */
```
ala-n, line 20:
```suggestion
  /** Default property value. Used if no attribute is present on the element. Empty string by default. Supports provider function. */
```

ala-n (approved):

NastaLeo (approved):

abarmina (approved):

dshovchko (approved):

PR #2195, opened 2024-01-25
fix(microtask): fix argument list on next fn call

nattallius:
Closes: 

Relates to:

### DIFF
--- a/src/modules/esl-utils/async/test/microtask.test.ts
+++ b/src/modules/esl-utils/async/test/microtask.test.ts
@@ -14,16 +14,23 @@ describe('sync/microtask', () => {
     await Promise.resolve();
     expect(fn).toBeCalledTimes(1);
   });
-  test('Decorated as microtask callback receives a correct list of call arguments', async () => {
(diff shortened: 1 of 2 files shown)


src/modules/esl-utils/async/test/microtask.test.ts
```diff
@@ -14,13 +14,18 @@ describe('sync/microtask', () => {
     await Promise.resolve();
     expect(fn).toBeCalledTimes(1);
   });
-  test('Decorated as microtask callback receives a list of call arguments', async () => {
+  test('Decorated as microtask callback receives a correct list of call arguments', async () => {
```
ala-n, line 17: Please split the cases.
Unit tests should be small, super straight, and simple, and ideally, we should not have multiple expectations from the single test test.
Test that `Decorated as microtask callback receives a list of call arguments` and then separately `Decorated as microtask callback refreshes after decorated method call (leak potected)`

ala-n (approved):

ala-n (approved):

abarmina (approved):

dshovchko (approved):

Issue #2153, opened 2024-01-08
[🚀site]: migrate `esl-tabs` example page to playground base


ala-n:
As an ESL site holder, I want to have the [`esl-tabs`](https://esl-ui.com/examples/tabs/) page to be migrated to the UIPlayground base.

The current page is **not proposed** to be saved as a draft page.

PR #2209, opened 2024-02-01
chore(site): ui-playground integration for tabs component

yadamskaya:
Closes: https://github.com/exadel-inc/esl/issues/2153


site/src/playground/export/lib.less
```diff
@@ -1,5 +1,4 @@
 @import '../../common/reboot.less';
-@import '../../common/typography.less';
```
ala-n, line 2: This will break the styles
I think we should use our typography in our examples.
If I understand correctly, that change relates to the default colors or padding, then I'd recommend 
  - for colors -> move them to the propper place (e.g. to `colors.less` instead of typography)
  - for padding -> just override them or move to `page.less` if this kind of changes required at all

dshovchko (approved):

ala-n (approved):

Issue #2218, opened 2024-02-05
[🚀esl-utils]: add attribute inheritance option for `@attr` decorator


ala-n:
As an ESL consumer, I want to have utilities to support common ESL component mechanics with an attribute property DOM-based inheritance.

It is proposed to extend `@attr` decorator with ability to declare `inherit` (named according discussion `1) attribute to inherit value:

1. Setter behavior does not affected, so `$el.myAttr = 'hi'` leads to own attribute mutation ignoring any inheritance logic.

2. Getter behavior should be updated according following cases:
    2.1. closest/inheritance feature is disabled initially, so no changes in existing `@attr` behavior
   
    2.2. if the element have a declared own attribute the closest/inheritance logic does not affect getter value. 
          In other words, one's own attribute has a priority over resolved by inheritance logic.
    
    2.3. when the `inherit` prop is declared with a boolean `true` value, the value is expected to be inherited from the closest parent attribute with the same name as current attribute declared.
      -  `@attr({ inherit: true }) public ignore;` -  finds `ignore` attr on this element or closest parent with attribute `ignore`
      -  `@attr({ inherit: true, dataAttr: true }) public ignore;` - finds `data-ignore` attr on this element or closest parent with  `data-ignore` attribute
   
    2.4. when the `inherit` prop is declared as a string, it is interpreted as an inherited attribute name to find the closest parent.  
      -  `@attr({ inherit: 'alt-ignore' }) public ignore;` - uses `ignore` attr on this element or closest parent with `alt-ignore` attribute
      -  `@attr({ inherit: 'alt-ignore', dataAttr: true }) public ignore;`  - uses `data-ignore` attr on this element or closest parent with `alt-ignore` attribute

ala-n:
Note/Question 1: the name of the attribute

- 👍  `closest`  ( `@attr({ closest: 'alt-ignore' }) public ignore;` )
- 🚀 `inherit`  ( `@attr({ inherit: 'alt-ignore' }) public ignore;` )
- 👎 - needs additional discussion for alt naming

CC: @NastaLeo , @exadel-inc/esl-core-maintainers

ala-n:
Note / Question 2: the logic is designed to use closed attribute finding, which includes the current element as a first target element.

```ts
   @attr({ closest: 'alt-ignore' }) public ignore;
```
```html
  <div alt-ignore="nope">
    <cust-el alt-ignore="hi"> // $0
  </div>
```

$0.ignore // hi

[...]

PR #2228, opened 2024-02-12
feat(esl-utils): extend `attr` decorator with inherit option

NastaLeo:
Closes: #2218

### DIFF
--- a/src/modules/esl-utils/decorators/attr.ts
+++ b/src/modules/esl-utils/decorators/attr.ts
@@ -15,11 +15,15 @@ type AttrDescriptor<T = string> = {
   name?: string;
   /** Create getter only */
   readonly?: boolean;
-  /** Find value to inherit across closest elements in DOM tree based on declared attribute name (in case of string format)
-   * or same attribute name of current element (boolean value).
-   * Example, attribute 'ignore' with configuration:
-   * inherit: 'alt-ignore' - searches ignore or data-ignore attr (in case of dataAttr: true) on this element or alt-ignore attribute on closest parent
-   * inherit: true - searches ignore or data-ignore attr (in case of dataAttr: true) on this element or on closest parent
+  /**
+   * Specifies the attribute inheritance behavior.
+   * If `inherit` is set to `true`, the attribute will inherit the value from the same-named attribute of the closest parent element in the DOM tree.
+   * For instance, `@attr({inherit: true}) ignore;` will look for an `ignore` attribute in the parent elements if it's not defined in the current element.
+   * If `dataAttr` is also true, it will search for `data-ignore` instead.
+   *
+   * If `inherit` is set to a string, it will use this string as the attribute name to search for in the parent elements.
+   * For example, `@attr({inherit: 'alt-ignore'}) ignore;` will first look for its own `ignore` attribute (or 'data-ignore' if `dataAttr` is true),
+   * and if not found, it will look for an `alt-ignore` attribute in the parent elements.
    */
   inherit?: boolean | string;
   /** Use data-* attribute */
--- a/src/modules/esl-utils/decorators/attr.ts
+++ b/src/modules/esl-utils/decorators/attr.ts
@@ -1,6 +1,6 @@
 import {identity, resolveProperty} from '../misc/functions';
 import {parseString, toKebabCase} from '../misc/format';
-import {getAttr, setAttr} from '../dom/attr';
+import {getAttr, getClosestAttr, setAttr} from '../dom/attr';
 
 import type {PropertyProvider} from '../misc/functions';
 import type {ESLAttributeDecorator} from '../dom/attr';
@@ -15,6 +15,13 @@ type AttrDescriptor<T = string> = {
   name?: string;
   /** Create getter only */
   readonly?: boolean;
+  /** Find value to inherit across closest elements in DOM tree based on declared attribute name (in case of string format)
+   * or same attribute name of current element (boolean value).
+   * Example, attribute 'ignore' with configuration:
+   * inherit: 'alt-ignore' - searches ignore or data-ignore attr (in case of dataAttr: true) on this element or alt-ignore attribute on closest parent
+   * inherit: true - searches ignore or data-ignore attr (in case of dataAttr: true) on this element or on closest parent
+   */
+  inherit?: boolean | string;
   /** Use data-* attribute */
   dataAttr?: boolean;
   /** Default property value. Used if no attribute is present on the element. Empty string by default. Supports provider function. */
@@ -36,14 +43,14 @@ const buildAttrName =
 export const attr = <T = string>(config: AttrDescriptor<T> = {}): ESLAttributeDecorator => {
   return (target: ESLDomElementTarget, propName: string): any => {
     const attrName = buildAttrName(config.name || propName, !!config.dataAttr);
+    const closestAttrName = typeof config.inherit === 'string' ? config.inherit : attrName;
 
     function get(): T | null {
-      const val = getAttr(this, attrName);
-      if (val === null && 'defaultValue' in config) {
-        return resolveProperty(config.defaultValue, this) as T;
-      }
+      const val = config.inherit ? getClosestAttr(this, closestAttrName) || getAttr(this, attrName) : getAttr(this, attrName);
+      if (val === null && 'defaultValue' in config) return resolveProperty(config.defaultValue, this) as T;
       return (config.parser || parseString as AttrParser<any>)(val);
     }
+
     function set(value: T): void {
       setAttr(this, attrName, (config.serializer as AttrSerializer<any> || identity)(value));
     }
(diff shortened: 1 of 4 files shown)


src/modules/esl-utils/decorators/attr.ts
```diff
@@ -15,6 +15,13 @@ type AttrDescriptor<T = string> = {
   name?: string;
   /** Create getter only */
   readonly?: boolean;
+  /** Find value to inherit across closest elements in DOM tree based on declared attribute name (in case of string format)
+   * or same attribute name of current element (boolean value).
+   * Example, attribute 'ignore' with configuration:
+   * inherit: 'alt-ignore' - searches ignore or data-ignore attr (in case of dataAttr: true) on this element or alt-ignore attribute on closest parent
+   * inherit: true - searches ignore or data-ignore attr (in case of dataAttr: true) on this element or on closest parent
+   */
```
ala-n, line 23: Just another interpretation, but still not clear enough. 
The main point - let's think about the description to make it more clear. 
```suggestion
  /** 
   * Inherits value from the attribute of the parent elements in the DOM tree. 
   * Uses the same attribute name as used for its own attribute if declared as `inherit: true`.
   * E.g., `@attr({inherit: true}) ignore;` - will search for `ignore` attribute on parents if its own value is not declared,
   * while `@attr({inherit: true, dataAttr}) ignore;` will look for `data-ignore` own or parent's attribute.
   * If the `inherit` value is declared as a string, then it considers the `inherit` parameter value as the closest attribute name to find the value to inherit.  
   * For example:
   * `@attr({inherit: 'alt-ignore'}) ignore;` - searches for its own `ignore` attribute (or 'data-ignore' in case of dataAttr: true) or consumes `alt-ignore` attribute (own or from the closest parent).
   */
```

src/modules/esl-utils/decorators/attr.ts
```diff
@@ -36,14 +43,14 @@ const buildAttrName =
...
   return (target: ESLDomElementTarget, propName: string): any => {
     const attrName = buildAttrName(config.name || propName, !!config.dataAttr);
+    const inheritAttrName = typeof config.inherit === 'string' ? config.inherit : attrName;
 
     function get(): T | null {
-      const val = getAttr(this, attrName);
-      if (val === null && 'defaultValue' in config) {
-        return resolveProperty(config.defaultValue, this) as T;
-      }
+      const val = config.inherit ? getClosestAttr(this, inheritAttrName) || getAttr(this, attrName) : getAttr(this, attrName);
```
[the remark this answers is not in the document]
ala-n: The question regarding esl-note can be considered as closed for now:
 - the logic should be inversed (second option) and we can not change it from user standpoint
 - the follow-up will be created for `esl-note` ignore property to replace it with visibility criteria or custom "inheritance"

src/modules/esl-utils/decorators/attr.ts
ala-n:
```suggestion
  /**
   * Specifies the attribute inheritance behavior. If `inherit` is set to `true`, the attribute will inherit the value from the same-named attribute of the closest parent element in the DOM tree. 
   * For instance, `@attr({inherit: true}) ignore;` will look for an `ignore` attribute in the parent elements if it's not defined in the current element. 
   * If `dataAttr` is also true, it will search for `data-ignore` instead.
   * 
   * If `inherit` is set to a string, it will use this string as the attribute name to search for in the parent elements. 
   * For example, `@attr({inherit: 'alt-ignore'}) ignore;` will first look for its own `ignore` attribute (or 'data-ignore' if `dataAttr` is true), and if not found, it will look for an `alt-ignore` attribute in the parent elements.
   */
```

src/modules/esl-utils/decorators/test/attr.test.ts
```diff
@@ -106,6 +107,117 @@ describe('Decorator: attr', () => {
...
+      test('The value resolves from the closest element (non-custom-element) in DOM', () => {
+        TEMPLATE.$firstEl.setAttribute('box', 'slide');
+        expect(TEMPLATE.$thirdEl.container).toBe('slide');
+      });
+      test('Elements with declared inherit are absent in DOM and returns empty string', () => {
+        expect(TEMPLATE.$thirdEl.container).toBe('');
+      });
+    });
+
+    describe('Inherit searches for the closest element with explicitly declared name', () => {
```
ala-n, line 162:
```suggestion
    describe('Inherit searches for the closest element with explicitly declared data attribute name', () => {
    // @attr({inherit: 'parent', dataAttr: true}) public ignore: string;
```

src/modules/esl-utils/decorators/test/attr.test.ts
```diff
@@ -106,6 +107,117 @@ describe('Decorator: attr', () => {
...
+      test('The value resolves from the closest element (non-custom-element) in DOM', () => {
+        TEMPLATE.$firstEl.setAttribute('parent', 'open');
+        expect(TEMPLATE.$thirdEl.ignore).toBe('open');
+      });
+      test('Elements with declared inherit are absent in DOM and returns empty string', () => {
+        expect(TEMPLATE.$thirdEl.ignore).toBe('');
+      });
+    });
+
+    describe('Inherit searches for the closest element with the same attribute name in DOM', () => {
```
ala-n, line 184:
```suggestion
    describe('Inherit searches for the closest element with the same attribute name in DOM', () => {
    // @attr({inherit: true}) public disallow: string;
```

src/modules/esl-utils/decorators/test/attr.test.ts
```diff
@@ -106,6 +107,117 @@ describe('Decorator: attr', () => {
...
+      test('The value resolves from the closest element (non-custom-element) in DOM with the same attribute name', () => {
+        TEMPLATE.$firstEl.setAttribute('disallow', 'deactivator');
+        expect(TEMPLATE.$thirdEl.disallow).toBe('deactivator');
+      });
+      test('Elements with declared inherit are absent in DOM and returns empty string', () => {
+        expect(TEMPLATE.$thirdEl.disallow).toBe('');
+      });
+    });
+
+    describe('Inherit searches for the closest element with the same data-attribute name in DOM', () => {
```
ala-n, line 202:
```suggestion
    describe('Inherit searches for the closest element with the same data-attribute name in DOM', () => {
    // @attr({inherit: true, dataAttr: true}) public allow: string;
```

ala-n (approved):

yadamskaya (approved):

src/modules/esl-utils/dom/test/attr.test.ts
```diff
@@ -118,4 +118,28 @@ describe('Attribute', () => {
...
+
+    const $parent = document.createElement('div');
+    const $parentName = 'parent-attr';
+    const $parentValue = 'parent-value';
+    $parent.setAttribute($parentName, $parentValue);
+
+    $parent.append($el);
+    document.body.append($parent);
+
+    test('finds indicated attribute on this element', () => {
```
dshovchko, line 135:
```suggestion
    test('should return attribute from the current element', () => {
```

src/modules/esl-utils/dom/test/attr.test.ts
```diff
@@ -118,4 +118,28 @@ describe('Attribute', () => {
...
+      expect(getClosestAttr($el, attrName)).toBe(attrValue);
+    });
+    test('finds indicated attribute on closest parent in DOM', () => {
```
dshovchko, line 138:
```suggestion
    test('should return an attribute from the parent DOM element', () => {
```

src/modules/esl-utils/dom/test/attr.test.ts
```diff
@@ -118,4 +118,28 @@ describe('Attribute', () => {
...
+      expect(getClosestAttr($el, $parentName)).toBe($parentValue);
+    });
+    test('returns null in case indicated attribute is absent in DOM', () => {
```
dshovchko, line 141:
```suggestion
    test('should return null in case the specified attribute is absent at the current element and its parents', () => {
```

src/modules/esl-utils/decorators/test/attr.test.ts
```diff
@@ -106,6 +107,121 @@ describe('Decorator: attr', () => {
...
+      //  @attr({inherit: 'box'}) public container: string;
+      test('The attribute inherit the value from own same-named attribute', () => {
+        TEMPLATE.$thirdEl.setAttribute('container', 'value');
+        expect(TEMPLATE.$thirdEl.container).toBe('value');
+      });
+      test('Own attribute setter changes the own value', () => {
+        TEMPLATE.$thirdEl.container = 'container';
+        expect(TEMPLATE.$thirdEl.container).toBe('container');
+      });
+      test('The value resolves from the closest element (custom-element) in DOM', () => {
```
dshovchko, line 150:
```suggestion
      test('The value resolves from the closest DOM element (custom-element)', () => {
```
or even
```suggestion
      test('The value should be resolved from the closest DOM element (custom-element)', () => {
```

dshovchko (approved):

Issue #2278, opened 2024-03-12
[🚀esl-utils]: add the ability to use additional attributes in loadScript utility


dshovchko:
As an ESL consumer, I want to be able to set additional attributes to the script tag when using the [loadScript()](https://github.com/exadel-inc/esl/blob/main/src/modules/esl-utils/dom/script.ts) utility. For example `crossorigin` or `referrerpolicy` attributes.

I propose to add a third argument to the `loadScript` method signature for passing additional attributes to set them at the script element.

```ts
loadScript('script-id', 'https://unpkg.com/foo.js', {crossorigin: 'anonymous', referrerpolicy: 'no-referrer'});
```

ala-n:
Note:
let's limit allowed settings to `crossorigin`, `referrerpolicy` and optionally `integrity` (not sure we need it, need to check if there is a possibility for such cases)
See comment in #2279 for details

PR #2279, opened 2024-03-12
add the ability to use additional attributes in loadScript utility

dshovchko:
Closes: #2278

### DIFF
--- a/src/modules/esl-utils/dom/script.ts
+++ b/src/modules/esl-utils/dom/script.ts
@@ -1,3 +1,5 @@
+import {setAttr} from './attr';
+
 export interface LoadScriptAttributes {
   crossorigin?: boolean | 'anonymous' | 'use-credentials' | '';
   integrity?: string;
@@ -12,7 +14,7 @@ const createScript = (id: string, src: string, attrs: LoadScriptAttributes): HTM
   const script = document.createElement('script');
   script.id = id;
   script.src = src;
-  Object.entries(attrs).forEach(([name, value]) => value && script.setAttribute(name, value === true ? '' : value));
+  Object.entries(attrs).forEach(([name, value]) => setAttr(script, name, value));
   return script;
 };
 
--- a/src/modules/esl-utils/dom/script.ts
+++ b/src/modules/esl-utils/dom/script.ts
@@ -1,9 +1,17 @@
+export interface loadScriptOptions {
+  crossorigin?: boolean | 'anonymous' | 'use-credentials' | '';
+  integrity?: string;
+  referrerpolicy?: ReferrerPolicy;
+}
+
 /**
  * Creates new script element with specified attributes
  */
-const createScript = (attrs: Record<string, string | boolean>): HTMLScriptElement => {
+const createScript = (id: string, src: string, options: loadScriptOptions): HTMLScriptElement => {
   const script = document.createElement('script');
-  Object.entries(attrs).forEach(([name, value]) => value && script.setAttribute(name, value === true ? '' : value));
+  script.id = id;
+  script.src = src;
+  Object.entries(options).forEach(([name, value]) => value && script.setAttribute(name, value === true ? '' : value));
   return script;
 };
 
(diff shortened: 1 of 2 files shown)


src/modules/esl-utils/dom/script.ts
```diff
@@ -1,23 +1,22 @@
...
-  script.src = src;
+  Object.entries(attrs).forEach(([name, value]) => value && script.setAttribute(name, value === true ? '' : value));
   return script;
 };
 
 /**
  * Common function that loads script async
  * @param id - unique script id that is used as a marker to prevent future load
  * @param src - script src (url) to load
+ * @param options - additional attributes to set on the script tag
```
ala-n, line 14: Mb lets limit the list of allowed parameters (at least from the types perspective but in that case i'd vote in general).
We are not providing utility to create script, the amount of capabilities should be strictly limited to the name an purpose of utility - load script file. So proposing to limit parmas list to cases we need in loading process :
 - crossorigin
 - referrerpolicy 
  - integrity (not sure we need)
 and probably that's it 
  We should not allow user to define useless attributes (defer, async, blocking, fetchprioriy, nonce, nomodeule - all of them will not work).
  `type` is also doesn't sound good here (we loading a simple old script, for loading modules user should consider browser capabilities to do that)

ala-n (changes requested):

src/modules/esl-utils/dom/script.ts
```diff
@@ -1,23 +1,22 @@
 /**
- * Creates new async script tag by id and src
+ * Creates new script element with specified attributes
  */
-const createAsyncScript = (id: string, src: string): HTMLScriptElement => {
+const createScript = (attrs: Record<string, string | boolean>): HTMLScriptElement => {
   const script = document.createElement('script');
-  script.id = id;
-  script.async = true;
```
ala-n, line 7: Note: the `async = true` - legacy operation here and mostly nominal (mb except IE). All currently supported browsers always load manually added scripts asynchronously.

dshovchko: ok, I will remove this

src/modules/esl-utils/dom/script.ts
ala-n: Sounds fine but I believe we are not gonna support anything additional. 

And small note for ideal typing:
replace strings with exact literals / types where it makes sense

E.g.:
```
   crossorigin?: boolean | 'use-credentials' | 'anonimus'; // Note: true = 'anonimus'
```

src/modules/esl-utils/dom/script.ts
```diff
@@ -1,23 +1,30 @@
+export interface loadScriptOptions {
+  crossorigin?: boolean | 'anonymous' | 'use-credentials' | '';
+  integrity?: string;
+  referrerpolicy?: ReferrerPolicy;
+}
+
 /**
- * Creates new async script tag by id and src
+ * Creates new script element with specified attributes
```
ala-n, line 8:
```suggestion
 * Creates new script element with specified attributes
 * Note: as the script will be added dynamically, it will be treated as async by the browser
```

src/modules/esl-utils/dom/script.ts
```diff
@@ -1,23 +1,30 @@
...
  */
-const createAsyncScript = (id: string, src: string): HTMLScriptElement => {
+const createScript = (id: string, src: string, options: loadScriptOptions): HTMLScriptElement => {
   const script = document.createElement('script');
   script.id = id;
-  script.async = true;
   script.src = src;
+  Object.entries(options).forEach(([name, value]) => value && script.setAttribute(name, value === true ? '' : value));
```
ala-n, line 14: Probably we can use setAttr from `dom/attr.ts`

ala-n (approved):

fshovchko (approved):

src/modules/esl-utils/dom/script.ts
dshovchko:
```ts
Object.entries(options).forEach(([name, value]) => setAttr(script, name, value));
```
I am not sure that it will be better because `setAttr()` is not short and simple. Also, it has additional dependency on `resolveDomTarget()`.

dshovchko: If do not take into view that someone will want to use this utility without using a base element or mixin element. But this looks strange.
So I will switch it to `setAttr()`.

NastaLeo (approved):

PR #2284, opened 2024-03-15
feat(esl carousel) rework css carousel

julia-murashko:


src/modules/esl-carousel/renderers/esl-carousel.css-only.renderer.less
```diff
@@ -1,60 +1,67 @@
...
     display: none;
-    width: 100%;
-    height: 100%;
+    backface-visibility: hidden;
+  }
+
+  esl-carousel-slide.forward,
+  esl-carousel-slide.backward,
+  esl-carousel-slide.right,
+  esl-carousel-slide.left {
```
ala-n, line 25: Use `animating` marker for main animation

src/modules/esl-carousel/renderers/esl-carousel.css-only.renderer.less
```diff
@@ -1,60 +1,67 @@
...
 
+  [data-slides-area] {
+    display: block;
+    position: relative;
+    width: 100%;
+    --offset: 0;
+    transform: translate3d(var(--offsetArea, 0), 0, 0);
+  }
+
+  &[animating] [data-slides-area] {
```
ala-n, line 12: In case this option conflicts with the main animation but is created for touch/drug support we need to consider separate marker for that one but not for main animation.

src/modules/esl-carousel/renderers/esl-carousel.css-only.renderer.ts
```diff
@@ -24,119 +26,102 @@ export class ESLCSSCarouselRenderer extends ESLCarouselRenderer {
...
+    $nextSlide.classList.add('next');
+
+    const dir = direction === 'next' ? 'right' : 'left';
+    $nextSlide.classList.add(dir);
+
+    $activeSlide.classList.add(to);
+    $nextSlide.classList.add(to);
 
-    // TODO: !
     return promisifyEvent(this.$carousel.$slidesArea, 'transitionend').catch(resolvePromise);
```
ala-n, line 63: 1) TODO: we need to leave some more options to catch this animation. E.g. make user able to use animation instead of transition.
2) It looks like it works (or may works incorrectly) from targeting perspective. We need to observe for transition of *specific properties* and for *slide itself*

3) We need to create good promisifyTransition/promisifyAnimation option + Note we have a global bug with non considering `transitioncancel` and the same problem in default renderer


In terms of current PR it is only required to have a noticeable TODO with mentioned undone factors

ala-n:
Mergin in bounds of CSS preparation

PR #2336, opened 2024-04-11
feat(esl-event-listener): separate intersection events

### LINKED ISSUE #2300: [🐛 esl-event-listener]:  `intersects` event with option `once: true` is being triggered even if target is not intersects viewport
**Input Code**
```js
export class MyCustomElement extends ESLBaseElement {
...
  @listen({
    event: 'intersects',
    target: ESLIntersectionTarget.for,
    once: true
  })
  _onIntersect(e: ESLIntersectionEvent): void {
    if (e.isIntersecting) {
      //do something
    }
  }
}
```

**Expected behavior/code**
`intersects` event should be triggered when target element intersects viewport first time

**Environment**
- Webpack 5.88.2
- Node/npm version: Node 18/npm 9
- Typescript: 5.2.2
- Target: ES6

fshovchko:
Closes #2300 

Not sure we need to handle intersection ratio cases, since we weren't handling it before. But maybe it needs to be changed.

### DIFF
--- a/src/modules/esl-event-listener/core/targets/intersection.event.ts
+++ b/src/modules/esl-event-listener/core/targets/intersection.event.ts
@@ -31,7 +31,7 @@ export class ESLIntersectionEvent extends Event implements IntersectionObserverE
   /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/IntersectionObserverEntry/time) */
   public readonly time: DOMHighResTimeStamp;
 
-  protected constructor(target: Element, eventName: ESLIntersectionEventType) {
+  protected constructor(eventName: ESLIntersectionEventType, target: Element) {
     super(eventName, {bubbles: false, cancelable: false});
     overrideEvent(this, 'target', target);
   }
@@ -46,7 +46,7 @@ export class ESLIntersectionEvent extends Event implements IntersectionObserverE
     * @param entry - The intersection observer entry.
    */
   public static fromEntry(eventName: ESLIntersectionEventType, entry: IntersectionObserverEntry): ESLIntersectionEvent {
-    const event = new ESLIntersectionEvent(entry.target, eventName);
+    const event = new ESLIntersectionEvent(eventName, entry.target);
     const {
       boundingClientRect,
       intersectionRatio,
--- a/src/modules/esl-event-listener/core/targets/intersection.event.ts
+++ b/src/modules/esl-event-listener/core/targets/intersection.event.ts
@@ -7,9 +7,14 @@ export type ESLIntersectionEventType = typeof ESLIntersectionEvent.TYPE | typeof
  * Based on the {@link IntersectionObserverEntry} object.
  */
 export class ESLIntersectionEvent extends Event implements IntersectionObserverEntry {
+  /** Type of event that will be dispatched on viewport exit */
   public static readonly OUT = 'intersects:out';
+  /** Type of event that will be dispatched on viewport enter */
   public static readonly IN = 'intersects:in';
+  /** Type of event that will be dispatched on both viewport enter and exit */
   public static readonly TYPE = 'intersects';
+  /** @deprecated use {@link TYPE} instead */
+  public static readonly type = this.TYPE;
 
   /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/IntersectionObserverEntry/target) */
   public override readonly target: Element;
--- a/src/modules/esl-event-listener/core/targets/intersection.event.ts
+++ b/src/modules/esl-event-listener/core/targets/intersection.event.ts
@@ -5,7 +5,12 @@ import {overrideEvent} from '../../../esl-utils/dom/events/misc';
  * Based on the {@link IntersectionObserverEntry} object.
  */
 export class ESLIntersectionEvent extends Event implements IntersectionObserverEntry {
-  public static readonly type = 'intersects';
+  public static readonly OUT_TYPE = 'intersects:out';
+  public static readonly IN_TYPE = 'intersects:in';
+  public static readonly BASE_TYPE = 'intersects';
+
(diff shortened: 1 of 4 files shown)


src/modules/esl-event-listener/core/targets/intersection.event.ts
```diff
@@ -22,14 +27,25 @@ export class ESLIntersectionEvent extends Event implements IntersectionObserverE
...
+  protected constructor(target: Element, eventName: typeof ESLIntersectionEvent.type) {
+    super(eventName, {bubbles: false, cancelable: false});
     overrideEvent(this, 'target', target);
   }
 
-  /** Creates {@link ESLIntersectionEvent} from {@link ESLIntersectionEvent} */
-  public static fromEntry(entry: IntersectionObserverEntry): ESLIntersectionEvent {
-    const event = new ESLIntersectionEvent(entry.target);
+  /** Validates event name. Accepted values are 'intersects:out', 'intersects:in', 'intersects' */
+  public static isValidEventType(event: string): boolean {
```
ala-n, line 36:
```suggestion
  public static isValidEventType(event: string): event is ESLIntersectionEventType {
```

src/modules/esl-event-listener/core/targets/intersection.event.ts
```diff
@@ -22,14 +27,25 @@ export class ESLIntersectionEvent extends Event implements IntersectionObserverE
   /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/IntersectionObserverEntry/time) */
   public readonly time: DOMHighResTimeStamp;
 
-  protected constructor(target: Element) {
-    super(ESLIntersectionEvent.type, {bubbles: false, cancelable: false});
+  protected constructor(target: Element, eventName: typeof ESLIntersectionEvent.type) {
+    super(eventName, {bubbles: false, cancelable: false});
```
ala-n, line 31: The order of the original constructor is preferable when it makes sense.
```suggestion
  protected constructor(eventName: typeof ESLIntersectionEventType, target: Element) {
    super(eventName, {bubbles: false, cancelable: false});
```

src/modules/esl-event-listener/core/targets/intersection.event.ts
```diff
@@ -5,7 +5,12 @@ import {overrideEvent} from '../../../esl-utils/dom/events/misc';
  * Based on the {@link IntersectionObserverEntry} object.
  */
 export class ESLIntersectionEvent extends Event implements IntersectionObserverEntry {
-  public static readonly type = 'intersects';
+  public static readonly OUT_TYPE = 'intersects:out';
+  public static readonly IN_TYPE = 'intersects:in';
+  public static readonly BASE_TYPE = 'intersects';
```
ala-n, line 10: Oh we are having a mess in event types constants:
 - `ESLIntersectionEvent.type` (and I think the most Events in `esl-event-listener` module)
 - `ESLCarouselSlideEvent.BEFORE` and `ESLCarouselSlideEvent.BEFORE`  (w/o postfix)
 
 In terms of what will be the ideal, I'd say:
  1. I'd not expect any other contents than the constants of the event name; I'd vote for names without postfixes/prefixes (moreover, for us, it's an extra bytes in a bundle). So for that one `IN`, `OUT`, `TYPE` (or `BASE`)
  2. Regarding the case, you are more right to choose the upper case as we are dealing with constants.
  The most complex one:
  3. The name for the most general event ... I'd actually vote for just `TYPE`, but not sure at all
  
  The only thing that I'm sure with: in case we are replacing the `type` we should deprecate it.
  
  @exadel-inc/esl-core-maintainers, you probably have some preferences/ideas regarding naming for event name constants. From my side, it's just thoughts; the only thing we need for sure is a consistent approach.

ala-n (approved):

src/modules/esl-event-listener/core/targets/intersection.event.ts
```diff
@@ -22,14 +31,22 @@ export class ESLIntersectionEvent extends Event implements IntersectionObserverE
   /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/IntersectionObserverEntry/time) */
   public readonly time: DOMHighResTimeStamp;
 
-  protected constructor(target: Element) {
-    super(ESLIntersectionEvent.type, {bubbles: false, cancelable: false});
+  protected constructor(target: Element, eventName: ESLIntersectionEventType) {
```
ala-n, line 34: Cosmetic
```suggestion
  protected constructor(eventName: ESLIntersectionEventType, target: Element) {
```

src/modules/esl-event-listener/core/targets/intersection.event.ts
```diff
@@ -22,14 +31,22 @@ export class ESLIntersectionEvent extends Event implements IntersectionObserverE
...
+  public static isValidEventType(event: string): event is ESLIntersectionEventType {
+    return event === ESLIntersectionEvent.TYPE || event === ESLIntersectionEvent.IN || event === ESLIntersectionEvent.OUT;
+  }
+
+  /** Creates {@link ESLIntersectionEvent} from {@link ESLIntersectionEvent}
+    * @param eventName - Custom event name ('intersects' by default. Accepted values are 'intersects:out', 'intersects:in', 'intersects').
+    * @param entry - The intersection observer entry.
+   */
+  public static fromEntry(eventName: ESLIntersectionEventType, entry: IntersectionObserverEntry): ESLIntersectionEvent {
+    const event = new ESLIntersectionEvent(entry.target, eventName);
```
ala-n, line 49:
```suggestion
    const event = new ESLIntersectionEvent(eventName, entry.target);
```

ala-n (approved):

ala-n (approved):

abarmina (approved):

PR #2354, opened 2024-04-17
feat(esl-utils): add provider function support into `@prop` decorator

ala-n:


src/modules/esl-utils/decorators/prop.ts
```diff
@@ -1,11 +1,32 @@
...
+
+/** Builds own property setter */
+function setter(name: string, readonly?: boolean) {
+  if (readonly) return (): void => void 0;
+  return function (value: any): void {
+    Object.defineProperty(this, name, {
+      value,
+      writable: true,
+      configurable: true
+    });
```
dshovchko, line 26: Should it be enumerable?

src/modules/esl-utils/decorators/test/prop.test.ts
```diff
@@ -1,22 +1,108 @@
...
+      class TestElement extends HTMLElement {
+        @prop({value: ''})
+        @attr()
+        public field: string;
+      }
+      new TestElement();
+    }).toThrowError(/own property/);
+  });
+
+  describe('Overriding @attr works fine', () => {
```
dshovchko, line 97: This looks like an assertion, not a describe of the tests.
MB
```suggestion
  describe('Checking the @attr overriding', () => {
```

src/modules/esl-utils/decorators/test/prop.test.ts
```diff
@@ -54,100 +140,82 @@ describe('Decorator: @prop', () => {
...
-    test('original decorator should not be executed', () => {
+    test('Original decorator should not be executed', () => {
       const el = new TestElement();
+      el.field = true;
       expect(el.getAttribute('field2')).toBe(null);
     });
   });
 
-  describe('Overriding @jsonAttr', () => {
+  describe('Overriding @jsonAttr with @prop works fine', () => {
```
dshovchko, line 170: and here
```suggestion
  describe('Checking the @jsonAttr overriding', () => {
```

dshovchko (approved):

src/modules/esl-utils/decorators/test/prop.test.ts
ala-n: BTW: ideal test should have a Test Case name, so it is statement but not the status or action
So I'd say both of our versions are not descriptive enough. So let's keep that update for now.

src/modules/esl-utils/decorators/prop.ts
ala-n: That's a good (and tricky) question, by the way... 

We have an enumerable option for the original property, which is not a used characteristic that was broken in addition.
Note that even if we pass enumerable: true, we will not get it enumerable on the instance; we just make it such on the prototype.
So, I'm not sure we chose the right strategy to support it at all.

For instance, the override that we are watching for, I believe it's an even more realistic case to make it enumerable, but I don't think it's popular.

So let's create an issue to come back to that question...

src/modules/esl-utils/decorators/test/prop.test.ts
dshovchko: ok

abarmina (approved):

NastaLeo (approved):

Issue #2345, opened 2024-04-16
[🚀 site]: fix problems on popup demo page, wrap it into PG


ala-n:
It is requested to prepare an example http://localhost:3005/draft/popup-game/ live.
NOTE: we are not adding it publically (so the page is still going to stay under the draft tag under the current story)

Issues:
 - [ ] the drug move is freezing and has a delay in response
 - [ ] wheel move leads to area resize
 
 Target: `main` is applicable, but `main-beta` should be fine

PR #2357, opened 2024-04-18
chore(site): rework `esl-popup` game demo page using ui playground

fshovchko:
Closes: #2345


site/src/esl-popup/esl-d-popup-game.ts
```diff
@@ -42,14 +42,14 @@ export class ESLDemoPopupGame extends ESLBaseElement {
...
+  @listen({auto: false, event: 'mouseup', target: window})
   protected _onDragStop(): void {
     this.dragging = false;
     this.$$off(this._onDragging);
     this.$$off(this._onDragStop);
   }
 
-  @listen({auto: false, event: 'mousemove'})
+  @listen({auto: false, event: 'mousemove', target: window})
   @decorate(rafDecorator)
```
ala-n, line 53:
```suggestion
```

ala-n (commented):
Works better but still not ideal.

Scenario 1: garb and move the "rocket" fast to the separate corner, but keep cursor inside the area.

Current result: the "rocket" and cursor shifted from each other
Expected result: cursor does not change it's position relatively to "rocket" until bounds reached.

Scenario 2: 
Thes suggestion regarding raf make things a little bit more unstable even if popup is not opened.
Expected result: it should be super smooth, the user can have a tens of popups and hudreds of  triggest, we should not have any performance issues on one trigger with closed popup (It could be a popup issue but it looks like it's only demo problem, popup feels itself super fine from consumers side with beta release at that moment)

site/views/draft/popup-game.njk
```diff
@@ -23,8 +23,8 @@ aside:
     --_font-size: calc(min(var(--_height, 300px), var(--_width, 300px)) / 3);
     position: relative;
     display: inline-block;
-    height: var(--_height, 300px);
-    width: var(--_width, 300px);
```
ala-n, line 27: But do we need the `_onWheel` handler then?

ala-n (commented):
Btw

dshovchko:
> Works better but still not ideal.
> 
> Scenario 1: garb and move the "rocket" fast to the separate corner, but keep cursor inside the area.
> 
> Current result: the "rocket" and cursor shifted from each other Expected result: the cursor does not change its position relatively to "rocket" until bounds are reached.
> 
> Suggestion: Please note that the delta approach is not stable for such cases in general. It's better to translate the total pointer movement to our coordinate system. Please look at the carousel touch plugin or scrollbar for reference.
> 
> Scenario 2: The suggestion bellow, regarding RAF decoration makes things a little bit more unstable even if the popup is not opened. Expected result: it should be super smooth, the user can have tens of popups and hundreds of triggest, and we should not have any performance issues on one trigger with a closed popup (It could be a popup issue but it looks like it's only demo problem, popup feels itself super fine from consumers side with beta release at that moment)

site/views/draft/popup-game.njk
dshovchko: I needed the opportunity to resize the container to check different cases with available space to position popup. The best decision that I think it up was the resizeable container. But I agree that it should be removed from this component into a mixin like `wheel-resiable-container` or something like it.

site/src/esl-popup/esl-d-popup-game.ts
dshovchko: Maybe this video - [I never thought of using CSS animations like this before!](https://www.youtube.com/watch?v=IdxzJLQ3Mbs) can be usefull as source of ideas for animations?

ala-n:
Let's discuss on the nearest ESL meeting

site/views/draft/popup-game.njk
```diff
@@ -66,108 +60,108 @@ aside:
...
+  <div class="col-12">
+    <uip-root>
+      <script type="text/html"
+              label="Horizontal"
+              uip-snippet
+              uip-snippet-js="js-snippet-popup-game">
+        <esl-d-popup-game trigger="::find(.game-popup-trigger)">
+          <esl-trigger
+            target="#game-popup"
+            mode="toggle"
```
ala-n, line 103:
```suggestion
            track-click="not" 
            track-hover
            hover-hide-delay="250"
```

`toggle` is a default
the hover behavior looks more suitable for demo purpose here

site/views/draft/popup-game.njk
```diff
@@ -66,108 +60,108 @@ aside:
...
+          document.querySelector('.popup-g-text').setAttribute('paragraphs', collapsed ? '.03' : '1');
+        }
+
+        document.querySelector('uip-root').addEventListener('uip:change', () => onUIPChange());
+
+        async function onUIPChange() {
+          await Promise.resolve();
+          document.querySelector('.popup-g-continue-btn').addEventListener('click', () => _onClick());
+          document.querySelector('.popup-g-hide-btn').addEventListener('click', () => _onClick());
+        }
```
ala-n, line 144: First of all, It's not obvious to the user, so I'd suggest avoiding scripts that user can not see in playground.

Second, not sure I like the idea of having custom scripts at all sее а couple of suggestions above.

site/views/draft/popup-game.njk
```diff
@@ -66,108 +60,108 @@ aside:
...
+          container=".uip-preview-inner"
+          dir="ltr"
+          margin-arrow="35">
+          <div class="popup-content-wrapper">
+            <div collapsed class="popup-content">
+              <a class="popup-g-hide-btn" href="javascript:void(0)" role="button">Collapse</a>
+              <lorem-ipsum class="popup-g-text" paragraphs=".03"></lorem-ipsum>
+              <a class="popup-g-continue-btn" href="javascript:void(0)" role="button">Expand</a>
+            </div>
+          </div>
```
ala-n, line 121:
```suggestion
          <div class="m-2" style="max-width: 500px">
            <p class="popup-content-wrapper">
              Space exploration has always captured the imagination of humanity. <esl-trigger target="::next" class="btn-link expand-trigger" active-class="hide">Expand</esl-trigger>
              <esl-toggleable>
                Rockets, the vehicles propelling us beyond our planet, are marvels of engineering and science.
                The journey begins with a fiery launch, overcoming Earth's gravity.
                As rockets ascend, they stage-separate, shedding weight to become more efficient.
                Reaching space, they enable satellites to orbit, telescopes to peer into the cosmos, and missions to distant planets.
                <esl-trigger target="::parent" class="btn-link">Collapse</esl-trigger>
              </esl-toggleable>
            </p>
          </div>
```

ala-n: For sure we need to cleanup inline styles above, limit them as much as we can, and include into visible part of playground

site/views/draft/popup-game.njk
```diff
@@ -12,162 +12,102 @@ aside:
...
+        </esl-popup>
+
+        <style>
+          .popup-content {
+            display: flex;
+          }
+          .popup-v .popup-content {
+            writing-mode: vertical-rl;
+            min-width: 50px;
+            height: 250px;
```
ala-n, line 70:
```suggestion
            min-height: 250px;
```

ala-n (approved):

site/views/draft/popup-game.njk
```diff
@@ -12,162 +12,102 @@ aside:
...
+                  Private companies like SpaceX, Blue Origin, and others are revolutionizing space travel with reusable rockets, reducing the cost of access to space and paving the way for potential future missions to Mars and beyond.
+                  Recent missions to Mars, such as NASA's Perseverance rover and China's Tianwen-1, aim to uncover the mysteries of the Red Planet and search for signs of past or present life.</p>
+              </details>
+          </div>
+          <span class="esl-popup-arrow"></span>
+        </esl-popup>
+
+        <style>
+          .popup-content {
+            display: flex;
```
dshovchko, line 65:
```suggestion
            display: flex;
            background: inherit;
```

dshovchko (approved):

NastaLeo (approved):

PR #2367, opened 2024-04-24
feat(esl-carousel): add handler for 'esl-show-request'

NastaLeo:


src/modules/esl-carousel/core/esl-carousel.ts
```diff
@@ -306,6 +306,15 @@ export class ESLCarousel extends ESLBaseElement {
     return el && el.hasAttribute('prev');
   }
 
+  // Navigates to requested slide
+  @listen('esl:show:request')
+  protected onHandleFootnotes(e: CustomEvent): void {
+    const target = (e.target as HTMLElement).closest('[esl-carousel-slide]') as HTMLElement;
```
ala-n, line 312: It looks good, but not perfect, just a couple of things we should consider here:

1) We have a really rare but realistic case when we have a nested carousel.
So in that case we will navigate trough the child carousel but we will not affect top parent carousel as the  `(e.target as HTMLElement).closest('[esl-carousel-slide]')` will give us only the most nested slide which is not member of parent carousel slide list.
We can discuss a solution for that case within the bounds of the current PR, or create that as a separate issue (If we really should harry up).

2) We had the support of the ` if (!isMatches(this, detail.match)) return null;` filter in toggleable, so we need to be consistent here as well.

src/modules/esl-carousel/core/esl-carousel.ts
```diff
@@ -306,6 +306,15 @@ export class ESLCarousel extends ESLBaseElement {
...
+    const index = this.$slides.indexOf(target);
+    if (index === -1) return;
+    this.goTo(`slide:${index}`);
```
ala-n, line 315: Should be fine unless we have a bug
```suggestion
    if (index !== -1) this.goTo(index);
```

ala-n:
Added `require squash` just in purpose to clarify public release note message a bit

src/modules/esl-carousel/core/esl-carousel.ts
```diff
@@ -199,6 +205,14 @@ export class ESLCarousel extends ESLBaseElement {
     this.renderer && this.renderer.redraw();
   }
 
+  @listen('esl:show:request')
+  protected onHandleFootnotes(e: CustomEvent): void {
```
dshovchko, line 209: There is a confusion with the name
Probably should be

```suggestion
  protected onShowRequest(e: CustomEvent): void {
```

dshovchko (approved):

Issue #2333, opened 2024-04-10
[esl-carousel] replace default arrow implementation with  a `clip-path`


ala-n:
Using `clip-path` to render arrows is shorter. In this case, carousel consumers can also easily change the color of the arrows depending on the carousel's state.

_Originally posted by @dshovchko in https://github.com/exadel-inc/esl/pull/2185#discussion_r1559277511_

PR #2368, opened 2024-04-24
refactor(esl-carousel): replace default arrow implementation with a clip-path

NastaLeo:
Closes: #2333

### DIFF
--- a/src/modules/esl-carousel/core/esl-carousel.less
+++ b/src/modules/esl-carousel/core/esl-carousel.less
@@ -5,8 +5,8 @@ esl-carousel {
   max-width: 100%;
   // Important to ignore padding/margin hack
   box-sizing: content-box;
-  margin: calc(var(--esl-carousel-clip-shift, 0) * -1);
-  padding: var(--esl-carousel-clip-shift, 0);
+  margin: calc(var(--esl-carousel-side-space, 0) * -1);
+  padding: var(--esl-carousel-side-space, 0);
 
   overflow: hidden; // Fallback
   /* stylelint-disable-next-line */
@@ -50,6 +50,6 @@ esl-carousel {
 }
 
 [esl-carousel-container] {
-  --esl-carousel-clip-shift: 5px;
+  --esl-carousel-side-space: 5px;
   position: relative;
 }
--- a/src/modules/esl-carousel/core/esl-carousel.less
+++ b/src/modules/esl-carousel/core/esl-carousel.less
@@ -5,8 +5,8 @@ esl-carousel {
   max-width: 100%;
   // Important to ignore padding/margin hack
   box-sizing: content-box;
-  margin: -5px;
-  padding: 5px;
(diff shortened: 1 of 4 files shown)


src/modules/esl-carousel/plugin/nav/esl-carousel.nav.arrows.less
```diff
@@ -28,11 +24,9 @@
   // Positioning
   .esl-carousel-arrow {
     width: var(--esl-carousel-arrow-size);
-    height: 100%;
+    height: var(--esl-carousel-arrow-size);
```
ala-n, line 27: The only case we need to consider for consumers (not implement, we need just keep it in mind): 
There is a case where we want to have a clickable area bigger than the arrow. What should we suggest in that case?

src/modules/esl-carousel/plugin/nav/esl-carousel.nav.arrows.less
```diff
@@ -1,15 +1,11 @@
 /** Arrow navigation styles */
 [esl-carousel-container] {
   --esl-carousel-arrow-size: 40px;
-  --esl-carousel-arrow-offset: clamp(
-    -40px,
-    calc(40px - 40 * 100% / 700),
-    -25px
-  );
+  //add 5px carousel padding
```
ala-n, line 4: Oh, I didn't think it's appear here )
Then we might need to declare a custom variable for our extended visibility area.

ala-n (approved):

src/modules/esl-carousel/core/esl-carousel.less
```diff
@@ -50,5 +50,6 @@ esl-carousel {
 }
 
 [esl-carousel-container] {
+  --esl-carousel-clip-shift: 5px;
```
dshovchko, line 53: What does it mean - clip-shift? I can't understand.
Maybe should name it `esl-carousel-arrow-shift`?

dshovchko (approved):

src/modules/esl-carousel/core/esl-carousel.less
ala-n: It's not related to the arrow, its the area around the carousel where overflow:clip does not have effect.
So we see how slides move and that area, an d also we use such area to display slide decorations like shadow, etc (otherwise overflow:clip will cut it of)

But agree that the variable name is not an ideal one, so any ideas welcome

dshovchko: I see. Maybe it should be `esl-carousel-side-margin`?

ala-n: Current vote for `esl-carousel-side-space`

Vote history:
esl-carousel-
  - side-space 1 1 1.5 1.5 2 1.5
  - side-offset 1 1 1.5 1.5 1.5
  - space-around 1 1.5 1

ala-n (approved):

abarmina (approved):

Issue #2390, opened 2024-05-07
[🚀esl-utils]: add ability to reject PromisifyEvent by using AbortSignal


dshovchko:
As a consumer of ESL PromisifyEvent, I want to have the ability to reject Promise by using the AbortSignal interface. This requested feature can assist in building complex asynchronous actions in the right manner.

Expected usage looks like that:

```ts
const controller = new AbortController();
const promise$ = promisifyEvent(el, 'click', null, {signal: controller.signal});
/*
    there is some useful code here
*/
if (somethingHappens) {
  controller.abort();
  // and now promise$ should be rejected
}
```

PR #2391, opened 2024-05-07
feat(esl-utils): add ability to reject promisifyEvent by using AbortSignal

dshovchko:
Closes: #2390


yadamskaya (approved):

src/modules/esl-utils/async/promise/event.ts
```diff
@@ -33,6 +33,9 @@ export function promisifyEvent(
     if (typeof timeout === 'number' && timeout >= 0) {
       setTimeout(eventCallback, timeout);
     }
+    if (typeof options === 'object') {
+      options.signal?.addEventListener('abort', () => reject(new Error('Rejected by abort signal')), {once: true});
+    }
```
ala-n, line 38: Memory leak: we have 2 cases when we end with signal subscription:
1) Negative (when original event observation is canceled by signal) - in that case we have an `abort` event and once will cover our requirement
2) Positive (the original Event was caught, so no signal involvement before event) - and here we end up with useless signal object that can not be garbage-collected as it have an active subscribtion.

src/modules/esl-utils/async/test/promise/event.test.ts
```diff
@@ -20,11 +20,12 @@ describe('async/promise/event', () => {
...
     });
 
-    test('Rejected by timeout if it is exceeded', () => {
+    test('Rejected by timeout if it is exceeded', async () => {
       const el = document.createElement('div');
       const promise$ = promisifyEvent(el, 'test', 10);
       jest.advanceTimersByTime(100);
-      return expect(promise$).rejects.toBeInstanceOf(Error);
+      await expect(promise$).rejects.toBeInstanceOf(Error);
+      await expect(promise$).rejects.toThrow(new Error('Rejected by timeout'));
```
ala-n, line 28:
```suggestion
      await expect(promise$).rejects.toThrow(new Error('Rejected by timeout'));
```

src/modules/esl-utils/async/test/promise/event.test.ts
```diff
@@ -37,6 +38,58 @@ describe('async/promise/event', () => {
...
       }
     });
+    describe('AbortSignal handling', () => {
+      test('Rejected when was passed options with signal and invoked abort', async () => {
+        const el = document.createElement('div');
+        const controller = new AbortController();
+        const promise$ = promisifyEvent(el, 'test', null, {signal: controller.signal});
+        controller.abort();
+        await expect(promise$).rejects.toBeInstanceOf(Error);
+        await expect(promise$).rejects.toThrow(new Error('Rejected by abort signal'));
```
ala-n, line 48:
```suggestion
        await expect(promise$).rejects.toThrow(new Error('Rejected by abort signal'));
```

src/modules/esl-utils/async/test/promise/event.test.ts
```diff
@@ -37,6 +38,58 @@ describe('async/promise/event', () => {
...
+      });
+      test('Rejected when was passed options with signal in aborted state', async () => {
+        const el = document.createElement('div');
+        const promise$ = promisifyEvent(el, 'test', null, {signal: AbortSignal.abort()});
+        await expect(promise$).rejects.toBeInstanceOf(Error);
```
ala-n, line 53:
```suggestion
```

ala-n (approved):

NastaLeo (approved):

fshovchko (approved):

Issue #2375, opened 2024-04-30
[🚀esl-media]: replace fill-mode with a CSS object-fit option for 5.0.0


ala-n:
As we no longer support IE11 we can use the possibilities of CSS  object-fit property to positioning video

However, we need to have backward compatibility, so `esl-media[fill-mode="cover"]` or `esl-media[fill-mode="inscribe"]` should still lead to the same visual result.

PR #2397, opened 2024-05-10
refactor(esl-media): `fill-mode` replaced with `object-fit` css property

fshovchko:
Closes: #2375


ala-n (changes requested):
Propose to consider one of the folowing technical approaches:

A.  
 - Declare `fitMode` as an official provider option.
 - Does not do anything in esl-media component
 - Implement helper polifill in a separate file or on the Base provider level (basically just resize callback)
 - Make subscription on the provide level
 - Deal with CSS for video object-fit trough the CSS (JS copy as it right now is ok but not perfect in terms of overriding and observing the fitMode property out of the box)
 
 B. Almost the same as option B
 But the resize subscription happens on esl-media to call abstract _onResize callback in case provider have 'isResizeAware' / `observeResize` marker.

src/modules/esl-media/core/esl-media-provider.ts
```diff
@@ -34,6 +34,7 @@ export type ProviderObservedParams = 'loop' | 'muted' | 'controls';
  */
 export abstract class BaseProvider {
   static readonly providerName: string;
+  static readonly isReplacable: boolean = false;
```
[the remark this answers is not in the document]
fshovchko: Decided to name flag accordingly to how `object-fit` applicable elements are named in spec: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

fshovchko:
> Propose to consider one of the folowing technical approaches:
> 
> A.
> 
> * Declare `fitMode` as an official provider option.
> * Does not do anything in esl-media component
> * Implement helper polifill in a separate file or on the Base provider level (basically just resize callback)
> * Make subscription on the provide level
> * Deal with CSS for video object-fit trough the CSS (JS copy as it right now is ok but not perfect in terms of overriding and observing the fitMode property out of the box)
> 
> B. Almost the same as option A But the resize subscription happens on esl-media to call abstract _onResize callback in case provider have 'isResizeAware' / `observeResize` marker.

Regarding `fitMode`. Yep, it would make sense to have an alias to esl-media's `fill-mode` on provider to be less dependent on esl-media. But we still need to know its `aspect-ratio` too. So i don't think there's much point in this introduction, and instead decided to stick with refresh method

src/modules/esl-media/providers/html5/video-provider.ts
```diff
@@ -12,6 +15,13 @@ export class VideoProvider extends HTMLMediaProvider {
 
   protected override _el: HTMLVideoElement;
 
+  public constructor(component: ESLMedia, config: MediaProviderConfig) {
+    super(component, config);
+    window.removeEventListener('resize', this.refreshElementSize);
+  }
+
+  protected override onResize(): void {}
```
ala-n, line 23: Doesn't look right
Providers should not subsctbe/unsubscribe or change anything outside from constructor

And I'm sure we can make resize subscription management more predictable then such side effect

ala-n (changes requested):
Temporary blocked. 
Require `condition` full support for listener for some further suggestions.

src/modules/esl-media/core/esl-media.less
```diff
@@ -50,5 +56,29 @@
...
+        height: 100%;
+        width: auto;
+      }
+  
+      &[fill-mode='inscribe'] .esl-media-inner {
+        height: auto;
+        width: 100%;
+      }  
+    }
   }
```
ala-n, line 83:
```suggestion

    &:is([fill-mode='cover'], [wide][fill-mode='inscribe']) .esl-media-inner {
      width: auto;
      height: 100%;
    }
    &:is([fill-mode='inscribe'], [wide][fill-mode='cover']) .esl-media-inner {
      width: 100%;
      height: auto;
    }
```

src/modules/esl-media/core/esl-media.ts
```diff
@@ -300,16 +302,9 @@ export class ESLMedia extends ESLBaseElement {
...
-      let stretchVertically = this.offsetWidth / this.offsetHeight < this.actualAspectRatio;
-      if (this.fillMode === 'inscribe') stretchVertically = !stretchVertically; // Inscribe behaves inversely
-      stretchVertically ?
-        this._provider.setSize(this.actualAspectRatio * this.offsetHeight, this.offsetHeight) : // h
-        this._provider.setSize(this.offsetWidth, this.offsetWidth / this.actualAspectRatio);   // w
-    } else {
-      this._provider.setSize('auto', 'auto');
-    }
+    const {actualAspectRatio} = this;
+    this.toggleAttribute('wide', this.offsetWidth / this.offsetHeight > actualAspectRatio);
```
ala-n, line 306:
```suggestion
    this.$attr('wide', this.offsetWidth / this.offsetHeight > actualAspectRatio);
```

src/modules/esl-media/core/esl-media.ts
```diff
@@ -300,16 +302,9 @@ export class ESLMedia extends ESLBaseElement {
...
-    if (this.fillModeEnabled && this.actualAspectRatio > 0) {
-      let stretchVertically = this.offsetWidth / this.offsetHeight < this.actualAspectRatio;
-      if (this.fillMode === 'inscribe') stretchVertically = !stretchVertically; // Inscribe behaves inversely
-      stretchVertically ?
-        this._provider.setSize(this.actualAspectRatio * this.offsetHeight, this.offsetHeight) : // h
-        this._provider.setSize(this.offsetWidth, this.offsetWidth / this.actualAspectRatio);   // w
-    } else {
-      this._provider.setSize('auto', 'auto');
-    }
+    const {actualAspectRatio} = this;
```
ala-n, line 305:
```suggestion
    if (!this._provider) return;
    const {actualAspectRatio} = this;
```
Still primary condition to do anything

src/modules/esl-media/core/esl-media-provider.ts
```diff
@@ -67,6 +67,10 @@ export abstract class BaseProvider {
     return this._ready;
   }
 
+  public set aspectRatio(ratio: number) {
+    this._el?.style.setProperty('aspect-ratio', ratio.toString());
+  }
+
```
ala-n, line 73: In case we have only one setter I'd recommend to place the exact set method nearby `setSize`

```
  public setAspectRatio(aspectRatio: number): void {
    this._el.style.setProperty('aspect-ratio', aspectRatio > 0 ? `${aspectRatio}`: null);
  }
```

src/modules/esl-media/core/esl-media.ts
```diff
@@ -300,16 +302,9 @@ export class ESLMedia extends ESLBaseElement {
...
+    this.toggleAttribute('wide', this.offsetWidth / this.offsetHeight > actualAspectRatio);
+    if (this._provider && this.fillModeEnabled && actualAspectRatio > 0) this._provider.aspectRatio = actualAspectRatio;
```
ala-n, line 307: We can set aspect-ratio independently of the this.fillModeEnabled
wide and height still primary metrics for CSS
aspectRatio just overrides the auto value for them

src/modules/esl-media/core/esl-media.less
fshovchko: > 

This suggestion is not logically correct.
For example: our video has wide aspect ratio and we want fill-mode=cover

Second selector [wide][fill-mode='cover']) fits since we have both attribute present.
First selector [fill-mode='cover'] fits since we have cover attribute present

First selector wins since it's first in the stylesheet file, but it wasn't supposed too. My suggestion is uglier, but it works

`
    &:not([wide])[fill-mode='cover'] .esl-media-inner, &[wide][fill-mode='inscribe'] .esl-media-inner {
      width: auto;
      height: 100%;
    }

    &:not([wide])[fill-mode='inscribe'] .esl-media-inner, &[wide][fill-mode='cover'] .esl-media-inner {
      width: 100%;
      height: auto;
    }

`

Please let me know if you have any other suggestions

src/modules/esl-media/core/esl-media.less
```diff
@@ -2,6 +2,12 @@
 .esl-media-init(@tagName: esl-media) {
   @{tagName} {
     display: block;
+    position: absolute;
+    inset: 0;
+    width: 100%;
+    height: 100%;
+    container-type: size;
+    overflow: hidden;
```
ala-n, line 10: It will definetly break something on consumers side
It looks like there is a custom styles but not the element base

abarmina (approved):

Issue #2393, opened 2024-05-07
[⚡e2e]: ability to store image snapshots as a WebP/AVIF


ala-n:
It is requested to investigate the capability of storing images snapshots in a webp/avif format to make it smaller.

ala-n:
Todo: @ala-n to close the issue

PR #2451, opened 2024-06-05
test(e2e): store images in webp format

fshovchko:
Closes: #2393


e2e/setup/scenarios.screenshot.ts
```diff
@@ -1,18 +1,33 @@
...
 import type {MatchImageSnapshotOptions} from 'jest-image-snapshot';
+import type {ScreenshotOptions} from 'jest-environment-puppeteer/node_modules/@types/puppeteer';
 
 const DIFF_CONFIG: MatchImageSnapshotOptions = {
   customDiffDir: '.diff'
 };
 
+async function pushWebPScreenshot(screenshots: (string | Buffer)[], options: ScreenshotOptions = {}): Promise<void> {
+  const screenshotBuffer = await page.screenshot(options);
+  const webpScreenshot = await sharp(screenshotBuffer).webp().toBuffer();
```
dshovchko, line 14: It seems to me that we should use `lossless=true` options here

ala-n:
It is definitely an enhancement that we can use, as we will get files that are practically smaller. 
However, it does not utilize the possibility of WebP compression. It improves the file size just because PNG compression that `sharp` uses by default is better compared to the raw PNG output that `pupiter` gives out of the box. So, we still store snapshots as PNGs, just with compression (a similar result can be achieved by using the sharp wrapper on top of the screenshot function `screenshots.push(await sharp(await page.screenshot(options)).png().toBuffer());`).

To summarize everything, I'd have to say:
1) The `sharp` library is definitely a good start, and we can leave it in the codebase with 100% confidence. (Just add a note in the README that it may require `npm install --include=optional sharp -g` to work properly in some cases.)
2) We can definitely transform this PR into a partial solution that provides compression, which is a valuable and valid improvement (just without WebP until it actually affects file format to store).
3) Regarding Dima's comment above, I'm not sure that lossy compression should be a problem unless it consistently gives us the same result. We are definitely not going to catch some color problems with the image, so unless compress(snapshot) = compress(actual) is valid for approximate similarity, we are fine.

4) Regarding the original ticket with WebP, here are some notes to move forward:
   - Snapshot storing is a `jest` part of the process.
   - It happens inside the matching library, specifically at the `toMatchImageSnapshot` level. If I remember correctly, we can describe to Jest itself how to store a snapshot, so we might not need to override `jest-image-snapshot` since it utilizes Jest's original snapshotting. But it still should be a little bit tricky thing to do as I mentioned initially.

dshovchko (commented):
If compression is ok for us I vote for storing images in AVIF format. AVIF has higher compression efficiency than WebP.

ala-n:
> If compression is ok for us, I vote for storing images in AVIF format. AVIF has higher compression efficiency than WebP.

This is a totally valid suggestion. If `shark` (or another image processor) can work with AVIF (approximately fast), we can use it. I'll update the original ticket.

e2e/package.json
```diff
@@ -15,11 +15,13 @@
   },
   "dependencies": {
     "@types/jest-environment-puppeteer": "^5.0.6",
-    "@types/jest-image-snapshot": "^6.4.0",
+    "@types/pixelmatch": "^5.2.6",
     "@types/puppeteer": "^7.0.4",
-    "jest-image-snapshot": "^6.4.0",
     "jest-puppeteer": "^10.0.1",
+    "looks-same": "^9.0.0",
```
ala-n, line 21: Do we still need it? Looks like we don't use it

e2e/setup/serializers/image/snapshot-matcher.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  protected async convertToRaw(input: string | Buffer): Promise<SharpContext> {
+    return sharp(input).ensureAlpha().raw().toBuffer({resolveWithObject: true});
+  }
+
+  protected async compareImages(prevImg: SharpContext, currImg: SharpContext): Promise<string | undefined> {
+    let diffPath: string | undefined;
+    const {width, height} = prevImg.info;
+    const diffBuffer = Buffer.alloc(width * height * 4);
+    const numDiffPixel = pixelmatch(prevImg.data, currImg.data, diffBuffer, width, height, this.config);
```
ala-n, line 68: TODO: check if it is possible to have an unhandled case here in case the snapshot is old or in incorrect format.
As far as I noticed pixelmatch throws an error of incomparable assets sizes.
We need to check if that possible and how it hadeled (should throw an snippet problem but not crash tests in general)

ala-n (approved):

ala-n:
The improvements work for me. The diff looks cool and catches major issues that we could have.
@exadel-inc/esl-core-team needs some check assistance from you.
In case everything is fine we can go with this implementation.

e2e/setup/serializers/image/snapshot-matcher.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (numDiffPixel > width * height * 0.0001) diffPath = await this.saveDiff(prevImg, currImg, diffBuffer);
+    return diffPath;
```
ala-n, line 76:
```suggestion
    if (numDiffPixel > width * height * 0.0001) return this.saveDiff(prevImg, currImg, diffBuffer);
```
Also cosmetic: Magic number `0.0001` should be in constant

e2e/setup/serializers/image/snapshot-matcher.ts
```diff
@@ -0,0 +1,94 @@
...
+  protected getMatcherResult(pass: boolean, message: string): jest.CustomMatcherResult {
+    return {pass, message: () => message};
+  }
+
+  protected async convertToRaw(input: string | Buffer): Promise<SharpContext> {
+    return sharp(input).ensureAlpha().raw().toBuffer({resolveWithObject: true});
+  }
+
+  protected async compareImages(prevImg: SharpContext, currImg: SharpContext): Promise<string | undefined> {
+    let diffPath: string | undefined;
```
ala-n, line 71:
```suggestion
```

e2e/setup/serializers/image/snapshot-matcher.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (!fs.existsSync(dirPath)) fs.mkdirSync(dirPath, {recursive: true});
+  }
+
+  protected getMatcherResult(pass: boolean, message: string): jest.CustomMatcherResult {
+    return {pass, message: () => message};
+  }
+
+  protected async convertToRaw(input: string | Buffer): Promise<SharpContext> {
+    return sharp(input).ensureAlpha().raw().toBuffer({resolveWithObject: true});
+  }
```
ala-n, line 68: Cosmetic:
```suggestion
  protected static async toRawImage(input: string | Buffer): Promise<SharpContext> {
    return sharp(input).ensureAlpha().raw().toBuffer({resolveWithObject: true});
  }
```

e2e/setup/serializers/image/snapshot-matcher.ts
```diff
@@ -0,0 +1,94 @@
...
+      await this.currentImg.toFile(prevImgPath);
+      return this.getMatcherResult(true, `New snapshot was created: ${prevImgPath}`);
+    }
+
+    const prevImg = await this.convertToRaw(prevImgPath);
+    const currImg = await this.convertToRaw(await this.currentImg.toBuffer());
+
+    let diffPath;
+    try {
+      diffPath = await this.compareImages(prevImg, currImg);
```
ala-n, line 49: It's an internal implementation, so it is a cosmetic, but  it is a little bit overmixed API for 
`compareImages` (in some cases we get error in some result)
I'd recommend to consider making it consistent - ideally safe-function :
 ```ts
 const diff = await this.compareImages(prevImg, currImg);
 
 if (!diff) return this.getMatcherResult(true, `Image is the same as the snapshot: ${prevImgPath}`);
 if (diff.reason = 'content') this.getMatcherResult(false, `Image mismatch found: ${diff.path}`);
 return this.getMatcherResult(false, `Error comparing snapshot to image ${prevImgPath}`);
 ```

e2e/setup/serializers/image/diff-builder.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export class DiffBuilder {
+
+  protected imgInfo: {width: number, height: number};
+
+  constructor(protected config: DiffBuilderConfig) {
+    this.imgInfo = this.config.img1.info;
+  }
+
+  public async build(): Promise<void> {
```
ala-n, line 19: Cosmetic (again that's internal implementation so no big concerns ):
The `build` function usually is immutable, while the current one have side effects (save to file)

Consider to rename class + function,
or execute final action outside class  
```ts

**Review by dshovchko — APPROVED** [review]

File: e2e/reporters/reporter.js [hunk 9]
```diff
@@ -48,14 +48,12 @@ class SnapshotAwareReporter {
     const filename = path.basename(testPath);
     const name = ancestorTitles.join(' > ');
     const statBase = {name, filename, status, title, time: duration};
-
     if (status === 'passed') return statBase;
 
-    const snapshotParts = [filename, ...ancestorTitles, title, '1-snap', 'diff'];
-    const snapshotName = snapshotParts.map(sanitize).join('-') + '.png';
+    const snapshotParts = [filename, ...ancestorTitles, title, 'diff'];
+    const snapshotName = snapshotParts.map(sanitize).join('-') + '.jpg';
```
ala-n, line 54: Please ensure you use totally the same test name builders inside the reporter and in the img serializer.
Sure it could be copy-passed for now (until it is written on different language platforms)

fshovchko: > Please ensure you use totally the same test name builders inside the reporter and in the img serializer. Sure it could be copy-passed for now (until it is written on different language platforms)

Speaking of different language platforms. Did some investigation, and ts script are still not supported

NastaLeo (approved):

e2e/reporters/reporter.js
ala-n: I'd suggest a separate module written on ES (JS) but with a ts declaration file, so you will be free to use both sources.

e2e/setup/serializers/image/image-processor.ts
```diff
@@ -30,7 +30,7 @@ export class SnapshotDataProcessor {
   public static async process(context: jest.MatcherContext, received: Buffer): Promise<SnapshotData> {
     const {testPath, currentTestName} = context;
     mkDir(SnapshotDataProcessor.SNAPSHOT_DIR);
-    const testName = toKebabCase(`${testPath!.split('/').pop()} ${currentTestName}`);
+    const testName = buildSnapshotName(testPath!.split('/').pop()!, currentTestName!);
```
ala-n, line 33: We talk about usage subdirecories including file name. 
So for
```
 tests
   subdir
     test.file.ts/girkin:
       - Test 1
       - Test 2
```
The output should be
```
 tests
   __snaoshots__
      subdir
        test.file
          test-1-snap.jpg
          test-2-snap.jpg
```
```
 .diff
    subdir
      test.file
        test-1-snap.jpg
        test-2-snap.jpg
```

e2e/reporters/reporter.js
```diff
@@ -48,14 +48,12 @@ class SnapshotAwareReporter {
     const filename = path.basename(testPath);
     const name = ancestorTitles.join(' > ');
     const statBase = {name, filename, status, title, time: duration};
-
     if (status === 'passed') return statBase;
 
-    const snapshotParts = [filename, ...ancestorTitles, title, '1-snap', 'diff'];
-    const snapshotName = snapshotParts.map(sanitize).join('-') + '.png';
+    const snapshotParts = [filename, ...ancestorTitles, title, 'diff'];
```
ala-n, line 53: As was mentioned before we use full test path and test full name (it is presented in the test object but probably it just called differently from the `currentTestName`, so that will be the only difference of how we pass input to blame builder)

ala-n, line 53: That cool for printing nut just one more highlight: 
 we don't use `ancestorTitles` to build the name, as it was not used in serializer
 we need exactly the same way to get file name

e2e/setup/serializers/image/utils/name.ts
```diff
@@ -0,0 +1,3 @@
+export const sanitize = (str: string): string => (str || '').replace(/\W+/g, '-').toLowerCase();
```
ala-n, line 1: Cosmetic:
rename/move/restructure: 
`e2e/serializers/image-snapshot.name.ts`
`e2e/serializers/image-snapshot.diff.ts`
`e2e/serializers/image-snapshot.pocessor.ts`

e2e/setup/serializers/image/diff-builder.ts
```diff
@@ -0,0 +1,42 @@
...
+import type {SharpContext} from './sharp';
+
+type DiffBuilderConfig = {
+  diffPath: string;
+  img1: SharpContext;
+  img2: SharpContext;
+  diffBuffer: Buffer;
+};
+
+export class DiffImageComposer {
```
ala-n, line 13: Smell (for future rework): the OOP approach (to be persize: creation of object composer that can not be reused but only saved sounds redundant) does not structure or give benefits to the purpose of the module.

e2e/setup/scenarios.screenshot.ts
```diff
@@ -29,14 +24,14 @@ cucumber.defineRule('take a screenshot of the element', async ({elements, screen
   screenshots.push(await element.screenshot());
 });
 
-cucumber.defineRule('check if the screenshot is exactly equal to the snapshotted version', ({screenshots}: TestEnv) => {
+cucumber.defineRule('check if the screenshot is exactly equal to the snapshotted version', async ({screenshots}: TestEnv) => {
   if (!screenshots.length) throw new Error('E2E: there is no any screenshot, make sure you have "Take a screenshot" before');
   const image = screenshots.pop();
-  expect(image).toMatchImageSnapshot(DIFF_CONFIG);
+  await expect(image).toMatchImageSnapshot();
```
ala-n, line 30: Is that aligned with jest concept, is that ok that we have async expect check?

fshovchko: > Is that aligned with jest concept, is that ok that we have async expect check?

Yes,
The reason for adding await, is because I've started to notice that error message would direct to the test one after the actual failed one

ala-n (changes requested):

e2e/reporters/reporter.js
```diff
@@ -44,22 +42,17 @@ class SnapshotAwareReporter {
   }
 
   buildTestStat(test, testPath) {
-    const {ancestorTitles, status, title, duration} = test;
+    const {status, title, duration, fullName} = test;
     const filename = path.basename(testPath);
-    const name = ancestorTitles.join(' > ');
```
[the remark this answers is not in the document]
fshovchko: Added `ancestoTitles` back, but we can use fullname instead, just to simplify. So far what i found is fullname is the same as `ancestorTitles` + `title` minus the formatting:

ancestor + title: _Feature: Homepage looks fine:test homepage footer on mobile_
fullName: _Feature: Homepage looks fine test homepage footer on mobile_

yadamskaya (approved):

abarmina (approved):

NastaLeo (approved):

ala-n (approved):

Issue #2377, opened 2024-04-30
[🚀esl-media-query]: (POC) consider out of the box support for esl-media-query-list syntax recognition


ala-n:
## Request
As for now we have two syntaxes to declare ESLMediaRuleList:
 - Tupple :   `ESLMediaRuleList.parse('value1|value2', '@mq1|@mq2');`
 - Arrow: `ESLMediaRuleList.parse('@mq1 => value1| @mq2 =>value2');`
 
 It is requested to have out of the box method to support both notations in case media tuple used as a default method.
 So 
 `ESLMediaRuleList.parse('@mqAlt1 => value1| @mqAlt2 =>value2', '@mq1|@mq2');` or
 `ESLMediaRuleList.parseAuto('@mqAlt1 => value1| @mqAlt2 =>value2', '@mq1|@mq2');` parse first argument as an arrow syntax without default tuple consideration.
 
  --- Usage Example ----
 ```
 /** Parses adaptive value to optimized {@link ESLMediaRuleList} */
export const parseAdaptiveValue = (value: string, mask: string): ESLMediaRuleList => {
  if (value.includes('=>')) return ESLMediaRuleList.parse(value);
  const length = value.split('|').length;
  mask= mask.split('|').slice(0, length).join('|');
  return ESLMediaRuleList.parse(mask, value);
};
 ```

---
 
## ACCETANCE CRITERIA 
 (Updated 16.07.2024)
  
 - Current `ESLMediaRuleList.parse` is deprecated and available for reimplementation in 5th version
 So current `parse` implementation is not requested to be saved
 
 - The result signature is expected to be the following:
 ```typescript
    /** Alias for parseQuery */
    parse(query: string): ESLMediaRuleList;
    /** Alias for parseQuery */
    parse(query: string, parser: RulePayloadParser): ESLMediaRuleList;
    
     /** Parses query considering it tuple or query syntax based on the actual value */
    parse(query: string, mask: string): ESLMediaRuleList;
     /** Parses query considering it tuple or query syntax based on the actual value with passed parser for values */
    parse(query: string, mask: string, parser: RulePayloadParser): ESLMediaRuleList;
 ```
 
 - USE DEFINITION OF THE ORIGINAL REQUEST AS A VALID AC

PR #2452, opened 2024-06-07
feat(esl-media-query): ingore tuple values if query syntax passed

fshovchko:
Closes: #2377

### DIFF
--- a/eslint/src/rules/4/deprecated.media-rule-list-parse.ts
+++ b/eslint/src/rules/4/deprecated.media-rule-list-parse.ts
@@ -8,11 +8,12 @@ import type {ESLintReplacementMethodCfg} from '../../core/deprecated-class-metho
 
 const AVAILABLE_SINCE = '5.0.0-beta.24';
 const isActual = lte(ESL_PACKAGE_VERSION, AVAILABLE_SINCE);
+const isTest = process.env.NODE_ENV === 'test';
 
 /**
  * Rule for deprecated 'parse' method of {@link ESLMediaRuleList}
  */
-export default isActual ?
+export default isActual || isTest ?
   buildRule({
     className: 'ESLMediaRuleList',
     deprecatedMethod: 'parse',
--- a/eslint/src/rules/4/deprecated.media-rule-list-parse.ts
+++ b/eslint/src/rules/4/deprecated.media-rule-list-parse.ts
@@ -8,12 +8,11 @@ import type {ESLintReplacementMethodCfg} from '../../core/deprecated-class-metho
 
 const AVAILABLE_SINCE = '5.0.0-beta.24';
 const isActual = lte(ESL_PACKAGE_VERSION, AVAILABLE_SINCE);
-const isTest = process.env.NODE_ENV === 'test';
 
 /**
  * Rule for deprecated 'parse' method of {@link ESLMediaRuleList}
  */
-export default isActual || isTest ?
+export default isActual ?
   buildRule({
     className: 'ESLMediaRuleList',
     deprecatedMethod: 'parse',
--- a/src/modules/esl-media-query/README.md
+++ b/src/modules/esl-media-query/README.md
@@ -204,20 +204,22 @@ ESLMediaRuleList.parse('1 | @XS => 2', String); // the same as sample above
 ESLMediaRuleList.parse('1 | @XS => 2', Number); // first query from the sample above that store numeric values
 ESLMediaRuleList.parse('@XS => {option: 1} | @+SM => {option: 2}', ESLMediaRuleList.OBJECT_PARSER); // second query from the sample above with an object payloads
 ESLMediaRuleList.parse('@XS => {option: 1} | @+SM => {option: 2}', evaluate); // the same as the sample above 
+ESLMediaRuleList.parse('1|2|3|4|5', '@xs|@sm|@md|@lg|@xl') // parses tuple payloads
 
 // Tupple parsing
 ESLMediaRuleList.parseTuple('@xs|@sm|@md|@lg|@xl', '1|2|3|4|5') // String payload example
 ESLMediaRuleList.parseTuple('@xs|@sm|@md|@lg|@xl', '1|2|3|4|5',  Number) // Numeric payload sample
 ```
-**Note**: Method `ESLMediaRuleList.parse` is deprecated, and will be reintroduced in ESL v5.0.0 with a different signature. For now use `ESLMediaRuleList.parseTuple` or `ESLMediaRuleList.parseQuery` instead.
 
 #### ESLMediaRuleList API
 
-- `ESLMediaRuleList.parse(ruleset: string)` - parse media ruleset defined with classic syntax mentioned in section above.
-Rules separated by `|` symbol, query and value separated by `=>` for each rule, query is optional.
+- `ESLMediaRuleList.parseTuple(queries: string, values: string)` - parse media ruleset from tuple of queries and values, all separated via `|` symbol.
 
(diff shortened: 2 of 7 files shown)


src/modules/esl-media-query/test/esl-media-rule-list.test.ts
```diff
@@ -178,4 +178,25 @@ describe('ESLMediaRuleList', () => {
       expect(() => ESLMediaRuleList.parseTuple('@xs', '1|2|3')).toThrowError();
     });
   });
+
+  describe('Mixed cases parsing', () => {
```
ala-n, line 182: Ideally is to focus on different input options here rather than on checking than check that media work (which is other test scope of work)

Please provide more cases. Id suggest the following at leat:


| 1-st arg | 2-nd arg | Result (we can use `toString` to check that everything parsed  correct) |
| --- | --- | --- |
| `123` | - | `all => 123` | 
| `(min-width: 100px) => 123` | - | `(min-width: 100px) => 123` |
| `0 \| (min-width: 100px) => 1 \|  (min-width: 200px) => 2` | - | `all => 0 \| (min-width: 100px) => 1 \|  (min-width: 200px) => 2` | 
| `0 \| (min-width: 100px) => 1 \|  (min-width: 200px) => 2` | `@xs \| @sm` | `all => 0 \| (min-width: 100px) => 1 \|  (min-width: 200px) => 2` | 
| `0 \| 1 ` | `@+xs \| @+sm` | ` (min-width: ...px) => 1 \| (min-width: ...px) => 2` |

ala-n:
The only open questions (for voting I guess) is the new method name and signature.

For the context:
The new method is used to parse query according to its syntax and based on user input so both
 `parse ('1 | 2 | 3', TUPPLE_MEDIA)` and `parse('@XS => 1 | @SM => 2 | @MD => 3', TUPPLE_MEDIA)` (TUPPLE_MEDIA provided but not used in that case) will be handled correctly.
 
 The purpose of creating this method appeared 
   - from the real project (where tupple with global media preset is used with potential usage of extended arrow syntax) 
   - and in purpose to extend carousel API 
     ```
       <esl-carousel media="@XS | @SM | @MD" count="1 | 2 | 3" ...>
       vs
       <esl-carousel count="1 | @SM => 2 | @MD => 3" ...>
     ```
     
Three is an existing method `parse` presented in ESLMediaRuleList but it can not be used to parse query in the mentioned way due to arguments order (array syntax and tupple media are the first arguments so it is mixed and depends on argument count)

1) So first of all we need a clear name for new method:
   a. `parseAddaptive` (currently in PR)
   b. `parseAuto`
   c. `autoparse`
   d. `autoParse`
   e. `parseAny`
   f. Your suggestion
   
2) The signature of the method is also a little bit questionable.
 I'd suggest to restrict it to 
 ```
 parseAdaptive(value: string, mask: string): ESLMediaRuleList;  
 parseAdaptive<U>(value: string, mask: string, parser: RulePayloadParser<U>): ESLMediaRuleList<U>;  
 ```
 
 We have a "low level" `parse` which accepts any argument count depending on what we are going to parse.
 The new method's purpose is always to parse uncertain user input, so providing media tuple is mandatory in that case.
 
 Please make me know if there is a concern with that point
 
 FYI: @exadel-inc/esl-all-contributors

eslint/src/rules/4/deprecated.media-rule-list-parse.ts
```diff
@@ -8,11 +8,12 @@ import type {ESLintReplacementMethodCfg} from '../../core/deprecated-class-metho
 
 const AVAILABLE_SINCE = '5.0.0-beta.24';
 const isActual = lte(ESL_PACKAGE_VERSION, AVAILABLE_SINCE);
+const isTest = process.env.NODE_ENV === 'test';
 
 /**
  * Rule for deprecated 'parse' method of {@link ESLMediaRuleList}
  */
-export default isActual ?
+export default isActual || isTest ?
```
ala-n, line 16: We probably can just disable that rule or exclude it from defalat at that point.

src/modules/esl-media-query/core/esl-media-rule-list.ts
```diff
@@ -35,64 +35,56 @@ export class ESLMediaRuleList<T = any> extends SyntheticEventTarget {
...
    */
-  public static parse<U>(mask: string, values: string, parser: RulePayloadParser<U>): ESLMediaRuleList<U>;
+  public static parse<U>(values: string, mask: string, parser: RulePayloadParser<U>): ESLMediaRuleList<U>;
   public static parse(query: string, ...common: (string | RulePayloadParser<any>)[]): ESLMediaRuleList {
     const parser: RulePayloadParser<any> = typeof common[common.length - 1] === 'function' ? common.pop() as any : String;
-    const value = common.pop();
-    return typeof value === 'string' ?
-      ESLMediaRuleList.parseTuple(query, value, parser) :
+    const mask = common.pop();
+    return typeof mask === 'string' ?
```
ala-n, line 81: It doesn't seems to be implements original ticket requiroments
We deprecated and exclude `parse` to reimplement it with "adaptive" logic to support both syntax based on past value but not just arguments list (that was in that method originally)

src/modules/esl-media-query/test/esl-media-rule-list.test.ts
```diff
@@ -8,7 +8,7 @@ describe('ESLMediaRuleList', () => {
 
   describe('Integration cases:', () => {
     test('Basic case: "1 | @sm => 2 | @md => 3" parsed correctly', () => {
-      const mrl = ESLMediaRuleList.parseQuery('1 | @sm => 2 | @md => 3');
+      const mrl = ESLMediaRuleList.parse('1 | @sm => 2 | @md => 3');
```
ala-n, line 11: Pleaes do not change existing tests. We do not change `parseQuery` or `parseTuple` and WWE don't gonna change their coverage.
We should test new parse method according to comments that we define for parseMixed that was here before...

src/modules/esl-media-query/README.md
```diff
@@ -204,20 +204,22 @@ ESLMediaRuleList.parse('1 | @XS => 2', String); // the same as sample above
...
 
-- `ESLMediaRuleList.parse(ruleset: string)` - parse media ruleset defined with classic syntax mentioned in section above.
-Rules separated by `|` symbol, query and value separated by `=>` for each rule, query is optional.
+- `ESLMediaRuleList.parseTuple(queries: string, values: string)` - parse media ruleset from tuple of queries and values, all separated via `|` symbol.
 
-- `ESLMediaRuleList.parseTuple(queries: string, values: string)` - parse media ruleset from tuple of
-queries and values, all separated via `|` symbol
+- `ESLMediaRuleList.parseQuery(ruleset: string)` - parse media ruleset defined with syntax, where query and value are a single argument separated by `=>` for each rule.
+
+- `ESLMediaRuleList.parse(values: string, queries: string)` - parse media ruleset defined with tuple syntax, where query and value are separate arguments.
```
ala-n, line 220: It doesn't seems to be correct according to request

ala-n (changes requested):

eslint/src/rules/4/deprecated.media-rule-list-parse.ts
fshovchko: > We probably can just disable that rule or exclude it from defalat at that point.

If we exclude it, do we to remove tests too?

ala-n: I'd say no, tests should be isolated and should not affect anything

ala-n (approved):

abarmina (approved):

dshovchko (approved):

NastaLeo (approved):

Issue #2361, opened 2024-04-20
[🚀es-lint]: add automatic warning when eslint-esl-plugin version is less then actual esl version


ala-n:
As an ESL consumer, I want to be notified when the ESLint plugin is not up to date. It is quite common for the @exadel/eslint-plugin-esl to be present in the project but not updated regularly.

ala-n:
Just a couple of technical notes regarding that request:

- What kind of mismatching is treated as incorrect?
   - A mismatch is considered incorrect when the version of the lint plugin is lower than the version of the library.

- Is the upper version of the lint plugin not compatible with the library?
  - No, it is compatible. One of the main use cases is for users to upgrade the lint plugin first before upgrading the ESL library.

- Can a major version mismatch be a problem?
  - It definitely could be, but it's not the focus of the check described in the current ticket. Major version compatibility will be handled explicitly with the peerDependency of the npm package.

- What is the main use case for the described warning feature?
  - The main use case is to recommend that users use an up-to-date plugin version during minor/patch library updates.

PR #2453, opened 2024-06-07
fix(lint): notify about version differences for `eslint-plugin-esl`

fshovchko:
Screenshot from 2024-06-10 17-36-47

Closes: #2361

### DIFF
--- a/eslint/src/core/check-version.ts
+++ b/eslint/src/core/check-version.ts
@@ -1,0 +1,25 @@
+import {lt} from 'semver';
+import color from 'kleur';
+
+const ESL_PACKAGE = '@exadel/esl';
+const PLUGIN_PACKAGE = '@exadel/eslint-plugin-esl';
+
+function getInstalledVersion(packageName: string): string | null {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+    return require(`${packageName}/package.json`).version;
+  } catch (error) {
+    return null;
+  }
+}
+
+export function checkVersion(): void {
+  const eslintVersion = getInstalledVersion(PLUGIN_PACKAGE);
+  const eslVersion = getInstalledVersion(ESL_PACKAGE);
+  if  (!(eslintVersion && eslVersion && lt(eslintVersion, eslVersion))) return;
+
+  console.log(`\n${color.yellow('⚠️ Warning:')}
+  Your installed version of ${color.yellow(PLUGIN_PACKAGE)} (${color.red(eslintVersion)}) \
+is lower than version of main package ${color.yellow(ESL_PACKAGE)} (${color.green(eslVersion)}).
+  Please update ${color.yellow(PLUGIN_PACKAGE)} to the latest version ${color.green(eslVersion)}\n`);
+}
--- a/eslint/package.json
+++ b/eslint/package.json
@@ -15,7 +15,9 @@
     "clean": "rimraf dist",
     "build": "npm run clean && tsc --project tsconfig.build.json"
   },
-  "dependencies": {},
+  "dependencies": {
+    "semver": "^7.6.2"
+  },
   "files": [
     "dist/**/*"
   ],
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -123,6 +124,7 @@
     "prettier": "^3.3.1",
     "rimraf": "^5.0.7",
     "semantic-release": "^24.0.0",
+    "semver": "^7.6.2",
```
ala-n, line 127: This is not good. While it's true that `semver` knows better how to compare versions, it's not a good idea to make the primary package depend on something required for another package and feature. I'm not even sure that we can place it in `devDependencies` as we are running it on the client side. (while placing in `dependecies` is not allowed for sure :) )

package.json
```diff
@@ -81,6 +81,7 @@
     "prepare": "husky && npm run build",
     "preversion": "npm test",
     "version": "cross-env-shell \"npm version $npm_package_version --no-git-tag-version --ws\" && git add **/package*.json",
+    "postinstall": "node build/eslint-esl.version.js",
```
ala-n, line 84: The idea is good. The primary case indeed starts from the ESL update – that is absolutely true. However, it doesn't seem 100% correct or optimal to make the ESL package worry about a separate optional package. It doesn't seem like it should be its job.

There are two options I can suggest to resolve the problem:

1. Use some completely indirect way to run a script located and operating within the `esl-eslint-plugin` from the `post-install` script of ESL. This way, we avoid explicit dependencies in ESL itself.

2. (This option sounds more suitable to me) Notify the user during the lint rule run itself. The message will not be missed in the upgrade log, as it will be posted each time ESLint runs with the ESL plugin.

ala-n (changes requested):

eslint/src/core/check-version.ts
```diff
@@ -0,0 +1,26 @@
+import {lt} from 'semver';
+import color from 'kleur';
+
+const eslintPackageName = '@exadel/eslint-plugin-esl';
+const eslPackageName = '@exadel/esl';
```
ala-n, line 5: Cosmetic, constant notation
```suggestion
const ESL_PACKAGE_NAME = '@exadel/esl';
const PLUGIN_PACKAGE_NAME = '@exadel/eslint-plugin-esl';
```
or just
```suggestion
const ESL_PACKAGE = '@exadel/esl';
const PLUGIN_PACKAGE = '@exadel/eslint-plugin-esl';
```

eslint/src/core/check-version.ts
```diff
@@ -0,0 +1,26 @@
...
+
+function getInstalledVersion(packageName: string): string | null {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+    const installedPackage = require(`${packageName}/package.json`);
+    return installedPackage.version;
```
ala-n, line 11: Cosmetic:
```suggestion
    return require(`${packageName}/package.json`).version;
```

ala-n (approved):

eslint/src/core/check-version.ts
```diff
@@ -0,0 +1,26 @@
...
+  } catch (error) {
+    return null;
+  }
+}
+
+export function checkVersion(): void {
+  const eslintVersion = getInstalledVersion(eslintPackageName);
+  const eslVersion = getInstalledVersion(eslPackageName);
```
fshovchko, line 19:
```suggestion
  const eslintVersion = getInstalledVersion(eslintPackageName);
  const eslVersion = getInstalledVersion(eslPackageName);
```
```suggestion
  const eslintVersion = getInstalledVersion(eslintPackageName);
  const eslVersion = getInstalledVersion(eslPackageName);
```
```suggestion
  const eslintVersion = getInstalledVersion(PLUGIN_PACKAGE);
  const eslVersion = getInstalledVersion(ESL_PACKAGE);
```

ala-n:
Maintainers note: change should be visible in release notes so consider `fix` or `MINOR VERSION` final commit message parts

NastaLeo (approved):

dshovchko (approved):

Issue #1954, opened 2023-09-25
[🐛 in esl-utils]: `debounce` method is incompatible with async void functions typescript 5


ala-n:
**Input Code**
```ts
  class Test {
    @decorate(debounce, 100) // TS: type are incompatible
    protected async update(): Promise<void> {
       // ... await
    }
  }
```

**Expected behavior/code**
Should wrap async method fine

**Environment**
- Webpack 5.88.2 + ts-lader 9.4.4
- Node 18; npm 8
- Typescript: 5.2.2
- Target: ES6

**Additional context**
Q: is it a valid situation ootb?

fshovchko:
After discussion with @exadel-inc/esl-core-team, it was decided to postpone the issue.
No simple solution was found, since altering types only would introduce additional issue.
Proposed solutions were considered too heavy and would probably require introduction of another parameter in `debounced` signature, or altering `debouncedSubject`

PR #2489, opened 2024-07-04
fix(esl-utils): debounced promise return (POC)

fshovchko:
Closes: #1954 
Request was to handle the following code
```
  class Test {
    @decorate(debounce, 100) // TS: type are incompatible
    protected async update(): Promise<void> {
       // ... await
    }
  }
```
In order to do so, debouncedSubject now needs to return the promise

### DIFF
--- a/src/modules/esl-utils/async/debounce.ts
+++ b/src/modules/esl-utils/async/debounce.ts
@@ -3,10 +3,14 @@ import {createDeferred} from './promise/defered';
 import type {AnyToAnyFnSignature} from '../misc/functions';
 import type {Deferred} from './promise/defered';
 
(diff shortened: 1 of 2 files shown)


src/modules/esl-utils/async/debounce.ts
```diff
@@ -37,7 +41,9 @@ export function debounce<F extends AnyToAnyFnSignature>(fn: F, wait = 10, thisAr
       deferred && deferred.resolve(result);
       deferred = null;
     }, wait);
+    return deferred.promise;
```
ala-n, line 44: Why do we change the actual return type?
It was made by request to make lazy creation of promise as the user does not need it in 95% of cases.

PR #2529, opened 2024-07-22
feat(esl-base-element): check if element is in the DOM on connected callback (POC)

fshovchko:
Recently we stumbled upon the following issue: _"[ESL]: No targets found for event listener"_

Upon investigation it was discovered that issue was appearing when multiple nested custom elements were re-appended to the DOM, with root element having logic that involves removal of the nested element from a child list, later causing duplicate connected callback and disconnected callback triggering, even though element wasn't technically appended to a DOM yet.

My guess that it might be that although element removal process is synchronous, it cannot remove element second time without re-appending it first.

### DIFF
--- a/site/views/draft/disconnected-callback.njk
+++ b/site/views/draft/disconnected-callback.njk
@@ -1,0 +1,22 @@
+---
+layout: empty
+title: Duplicated disconnected callback causing falsy duplicate connected callback
+name: Disconnected callback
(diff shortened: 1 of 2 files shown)


ala-n:
@ala-n to handle risks and client testing

ala-n:
Reopened in #2667

PR #2555, opened 2024-07-31
feat(esl-carousel): rework default renderer animation approach (now uses js animation)

ala-n:


fshovchko (approved):

src/modules/esl-carousel/renderers/esl-carousel.default.renderer.ts
```diff
@@ -72,18 +71,19 @@ export class ESLDefaultCarouselRenderer extends ESLCarouselRenderer {
...
-    if (!transform || transform === 'none') return 0;
-    const position = this.vertical ? 5 : 4; // tx or ty position
-    return parseInt(transform.split(',')[position], 10);
+  /** Animate scene offset */
+  protected async animateTransformOffset(offset: number = -this.getOffset(this.currentIndex)): Promise<void> {
+    this.$carousel.$$attr('animating', true);
+    await this.$area.animate({
+      transform: [`translate3d(${this.vertical ? `0px, ${offset}px` : `${offset}px, 0px`}, 0px)`]
+    }, {duration: 250, easing: 'linear'}).finished;
+    this.$carousel.$$attr('animating', false);
```
dshovchko, line 86: Just cosmetic to define a method something like `setAnimationState()`

dshovchko (approved):

src/modules/esl-carousel/renderers/esl-carousel.default.renderer.ts
ala-n: But what will be the purpose? it is not public and is commonly used for other methods or inheritors. From the current update state `animateTransformOffset` and only `animateTransformOffset` should process animation and notify regarding the animation state. It is not designed for animated markers to be updated out of the context of animation

abarmina (approved):

PR #2574, opened 2024-08-07
feat(esl-image-utils): create esl-img-container mixin to provide imgcontainer functionality

ala-n:


src/modules/esl-image-utils/README.md
```diff
@@ -0,0 +1,89 @@
...
+<div class="img-container">
+  <picture>
+    <source srcset="image.webp" type="image/webp">
+    <img alt src="image.jpg" esl-img-container="::closest(.img-container)"/>
+  </picture>
+</div>
+```
+Note: it is not necessary to use `::closest` query in current case. `::parent` query with selector `img-container` will work as well.
+
+Set custom class `loaded` on the direct parent of the image element upon image load.
```
NastaLeo, line 42: mb simplify examples with error and loaded classes?

site/views/examples/image-utils.njk
```diff
@@ -0,0 +1,64 @@
...
+          <option value="img-container-16-9">16 x 9</option>
+          <option value="img-container-4-3">4 x 3</option>
+          <option value="img-container-1-1">1 x 1</option>
+        </uip-select-setting>
+        <uip-select-setting target=".img-container img" label="Image mode:" attribute="class" mode="append">
+          <option value="">Stretch</option>
+          <option value="img-cover">Cover</option>
+          <option value="img-contain">Inscribe (Fit)</option>
+        </uip-select-setting>
+        <uip-select-setting target=".img-container :is(.img-cover, .img-inscribe)" label="Image alignment:" attribute="style" mode="append">
```
NastaLeo, line 52:
```suggestion
        <uip-select-setting target=".img-container :is(.img-cover, .img-contain)" label="Image alignment:" attribute="style" mode="append">
```

NastaLeo (approved):

src/modules/esl-image-utils/core/esl-image-conyainer.less
```diff
@@ -0,0 +1,44 @@
+// Image Container Defaults
+.img-container {
+  position: relative;
+  overflow: hidden;
+
+  img {
+    position: absolute;
+    top: 0;
```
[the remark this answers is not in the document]
ala-n: `bottom: auto; right: auto;`
So `inset: 0 auto auto 0;` which is not so short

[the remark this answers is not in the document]
NastaLeo: oops, I missed smth
thanks for explanation!

src/modules/esl-image-utils/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Set custom class `loaded` on the direct parent of the image element upon image load and class `error` in case of image loading error.
+```html
+<div class="img-container" esl-img-state-cls="loaded" esl-img-state-cls-error="error">
+  <img alt src="image.jpg"/>
+</div>
+```
+
+Set custom class `loaded` on the direct parent of the image element upon image successful load.
+Note: the error class query executed after the plain class query. So error class query could override the plain class query.
```
dshovchko, line 62:
```suggestion
Sets custom class `loaded` on the direct parent of the image element upon the image successfully loading.
Note: the error class query is executed after the plain class query. So error class query could override the plain class query.
```

dshovchko (approved):

ala-n:
Will be reopened when the final design will be discussed

PR #2670, opened 2024-09-19
test(e2e): handle sanpshot's size mismatch

fshovchko:
After test run of a new snapshot matcher implementation, it was discovered that reports became less informative.
The most important things are:
- Missing reports on when two snapshot have mismatching size
- Error messages are less informative

### DIFF
--- a/e2e/setup/serializers/image-snapshot.sharp.ts
+++ b/e2e/setup/serializers/image-snapshot.sharp.ts
@@ -21,4 +21,38 @@ export class SharpService {
   public static toJPEGBufferd(input: string | Buffer, options: any = {}): Promise<Buffer> {
     return SharpService.toJPEG(input, options).toBuffer();
   }
+
+  public static async normalizeImages(img1: Buffer | string, img2: Buffer | string): Promise<[Buffer, Buffer]> {
+    const imgCtx1 = sharp(img1);
+    const imgCtx2 = sharp(img2);
+
+    const metadata1 = await imgCtx1.metadata();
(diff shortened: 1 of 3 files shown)


ala-n (approved):

e2e/setup/serializers/image-snapshot.sharp.ts
```diff
@@ -21,4 +21,40 @@ export class SharpService {
...
+      if (metadata.width === width && metadata.height === height) return image.toBuffer();
+
+      const padX = Math.max(0, (width - metadata.width!) / 2);
+      const padY = Math.max(0, (height - metadata.height!) / 2);
+
+      return image.extend({
+        top: Math.floor(padY),
+        bottom: Math.ceil(padY),
+        left: Math.floor(padX),
+        right: Math.ceil(padX),
```
ala-n, line 50: TODO: Align to top/left

e2e/setup/serializers/image-snapshot.sharp.ts
```diff
@@ -21,4 +21,40 @@ export class SharpService {
...
+    const imgCtx1 = sharp(img1);
+    const imgCtx2 = sharp(img2);
+
+    const metadata1 = await imgCtx1.metadata();
+    const metadata2 = await imgCtx2.metadata();
+
+    const targetWidth = Math.max(metadata1.width!, metadata2.width!);
+    const targetHeight = Math.max(metadata1.height!, metadata2.height!);
+
+    const padImage = async (
```
ala-n, line 35: Separate method instead of anonymous function

Issue #2712, opened 2024-10-21
[🚀esl-media]: out of the box observartion for container state


ala-n:
As an ESL user, I could use esl-media inside containers based on esl-toggleable. I want to have out of the box control for container state in that case:
 - if the toggleable closed -> esl-media paused
 - if the toggleable opened -> esl-media with autoplay feature starts playing
   Note: make sure if the play-in-viewport specified video will not start playing outside of the viewport

PR #2720, opened 2024-10-22
feat(esl-media): observation for togglebale container state

fshovchko:
Closes: #2712


src/modules/esl-media/core/esl-media.ts
```diff
@@ -371,6 +374,26 @@ export class ESLMedia extends ESLBaseElement {
...
+    target: window
+  })
+  protected _onContainerShow(e: Event): void {
+    const {target} = e;
+    if (!this.isAcceptableTarget(target) || !this.inViewport) return;
+    if (this.autoplay) this.play();
+  }
+
+  @listen({
+    event: ESLToggleable.prototype.HIDE_EVENT,
```
ala-n, line 388: Probably we need to use string literal instead as currently, we obligate the user to have both components in the bundle.

src/modules/esl-media/core/esl-media.ts
```diff
@@ -381,6 +404,10 @@ export class ESLMedia extends ESLBaseElement {
     }
   }
 
+  protected isAcceptableTarget(target: EventTarget | null): boolean {
+    return isElement(target) && target.contains(this);
+  }
```
ala-n, line 409: Move to esl-utils/dom/traversing, please.
Use something like `isContain(container, element)`
PS: `isRelativeNode` could reuse `isContain` then

src/modules/esl-media/core/esl-media.ts
```diff
@@ -99,6 +100,8 @@ export class ESLMedia extends ESLBaseElement {
   @boolAttr() public playsinline: boolean;
   /** Allows play resource only in viewport area */
   @boolAttr() public playInViewport: boolean;
+  /** Viewport intersection marker */
+  @boolAttr() public inViewport: boolean;
```
ala-n, line 104:
```suggestion
  @boolAttr({readonly: true}) public inViewport: boolean;
```
+move to readonly accessors declararion

src/modules/esl-media/core/esl-media-iobserver.ts
```diff
@@ -27,11 +27,13 @@ function handleViewport(entry: IntersectionObserverEntry): void {
   }
 
   // Videos that playing and out of min ratio RATIO_TO_STOP should be stopped
-  if (video.active && entry.intersectionRatio <= RATIO_TO_STOP) {
-    video.pause();
+  if (entry.intersectionRatio <= RATIO_TO_STOP) {
+    if (video.active) video.pause();
+    video.$$attr('in-viewport', false);
```
ala-n, line 32: The biggest problem now is that we need to revisit this logic.
It doesn't sound correct that we affect component twicely: more indirect - attribute, directly - action (play/pause)
In case we going to maintain indirect state it sounds like we need to move activation logic on esl-media canPlay level.

ala-n (changes requested):

src/modules/esl-media/core/esl-media-iobserver.ts
fshovchko: > The biggest problem now is that we need to revisit this logic. It doesn't sound correct that we affect component twicely: more indirect - attribute, directly - action (play/pause) In case we going to maintain indirect state it sounds like we need to move activation logic on esl-media canPlay level.

Now we introduced way too many switch cases
`  protected override attributeChangedCallback(attrName: string, oldVal: string, newVal: string): void {
    if (!this.connected || oldVal === newVal) return;
    // eslint-disable-next-line sonarjs/max-switch-cases`

src/modules/esl-utils/dom/traversing.ts
```diff
@@ -1,3 +1,4 @@
+import {isElement} from './api';
```
ala-n, line 1: Replace import with a more specific one. Ideally the consumers should have tree-shaking but we should not rely on it

src/modules/esl-media/core/esl-media.ts
```diff
@@ -213,6 +220,14 @@ export class ESLMedia extends ESLBaseElement {
     return this.conditionQuery.matches;
   }
 
+  protected canPlay(): boolean {
+    return !this.playInViewport || (!this.active && this.autoplay && this.inViewport);
+  }
+
+  protected canPause(): boolean {
+    return this.active;
+  }
```
ala-n, line 229: It is strange to have this method we should not overcomplicate esl-media

src/modules/esl-media/core/esl-media.ts
```diff
@@ -213,6 +220,14 @@ export class ESLMedia extends ESLBaseElement {
     return this.conditionQuery.matches;
   }
 
+  protected canPlay(): boolean {
+    return !this.playInViewport || (!this.active && this.autoplay && this.inViewport);
+  }
```
ala-n, line 225: Here is a similar situation We haven't this check previously.
Can't we modify behaviour of existing canActivate ...

src/modules/esl-media/core/esl-media.ts
```diff
@@ -205,6 +209,9 @@ export class ESLMedia extends ESLBaseElement {
         this.$$on(this._onConditionChange);
         this.deferredReinitialize();
         break;
+      case 'in-viewport':
+        this.inViewport ? this.play() : this.pause();
```
ala-n, line 213: should also relay on autoplay allowance
Probaly we need a special method to handle (in-viewport behavior + esl:show/esl:hide related changes)
(As an idea we can use the same attribute to make esl-media handle both scenarios hide due to viewport leave or hide due to esl:hide event)

I wasn't just assume that we can have troubles with playInVIewport feature. It is the same feature in general, observation of toggleable state is just less restrictive version of play-in-viewport that should handle only (visibility state change independently of viewport)

src/modules/esl-media/core/esl-media.ts
```diff
@@ -335,6 +346,28 @@ export class ESLMedia extends ESLBaseElement {
...
   }
 
+  @listen({
+    event: 'esl:show',
+    target: window
+  })
+  protected _onContainerShow(e: Event): void {
+    const {target} = e;
+    if (!isSafeContains(target as Node, this)) return;
+    if (this.playInViewport && isVisible(this)) return;
```
ala-n, line 356: Probably here we should have the following condition
```suggestion
    if (this.playInViewport && !isVisible(this)) return;
```

src/modules/esl-media/core/esl-media.ts
```diff
@@ -81,6 +82,8 @@ export class ESLMedia extends ESLBaseElement {
   @attr({parser: parseLazyAttr, defaultValue: 'none'}) public lazy: ESLMediaLazyMode;
   /** Autoplay resource marker */
   @boolAttr() public autoplay: boolean;
+  /** Will prevent autoplay if media was paused manually */
+  @boolAttr() public noAutoplayOnInterrupt: boolean;
```
ala-n, line 86: Actually, my only big concern is the name and default for that variable...
Regarding default I propose invert of the option...
Regarding naming
   I'd consider special string value for autoplay (e.g. ``autoplay="always"` and doc note about its behaviour)
      
Another option is `autoplay-on-interrupt` but it is still long

@exadel-inc/esl-all-contributors feel free to jump with suggestions and comments here

src/modules/esl-media/core/esl-media.ts
```diff
@@ -335,6 +346,28 @@ export class ESLMedia extends ESLBaseElement {
     this.deferredReinitialize();
   }
 
+  @listen({
```
ala-n, line 349: Regarding listeners, we have here, it's perfect if we limit them to a single actual subscription (move to service, which will probably also handle observer file functionality)
But initially and for that pr I'm more than fine to have them as it is.

ala-n (commented):
Solutions looks logical to me.
Attribute `autopaused` looks super clear and logical to me...

PR #2725, opened 2024-10-28
test(esl-media): e2e testing

### LINKED ISSUE #2709: [⚡tests ]: make sure YouTube/HTML5 ESLMedia provides are covered by tests
As an ESL maintainer, I want to be sure ESLMedia providers/core is covered by tests (unit + potentially e2e)

fshovchko:
Closes: #2709

@ala-n, do we need any other functionality tested or basic check for video/iframe tag is enough?


ala-n:
Nope, checking the existence of an iframe is not enough.
From what I could suggest based on YouTube provider work  we need:
  1) Check that YouTube API is loaded, check that it is loaded once
  2) We need to check the inner iframe is, actually, a YouTube and that it at least contains an ID of the required video
  3) We need to check initial attributes as well (`playsinline`, `autoplay`, `mute`)
  4) Ideally we need as much as possible including the ability to play etc. but unfortunately, YouTube no longer allows playing videos for unauthorized users (autotests including).

e2e/tests/modules/esl-media/utils/utils.ts
```diff
@@ -0,0 +1,16 @@
+import type {ESLMedia} from '@exadel/esl/modules/esl-media/core/esl-media';
+import type {JSHandle} from 'puppeteer';
+
+export const createMediaElement = async (props: Partial<ESLMedia>): Promise<JSHandle<ESLMedia>> => {
+
+  return page.evaluateHandle((pr: Partial<ESLMedia>) => {
+    const toKebabCase = (str: string): string => str.replace(/([a-z])([A-Z])/g, '$1-$2').replace(/[\s_]+/g, '-').toLowerCase();
```
ala-n, line 7: Probably could be replaced with ESL `toKebabCase`

e2e/tests/modules/esl-media/utils/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    const $media = document.createElement('esl-media');
+    Object.entries(pr).forEach(([attr, value]) => {
+      $media.setAttribute(toKebabCase(attr), value?.toString() || '');
```
ala-n, line 11: Propbaly in case ESLMedia is already registered you can just do

`Object.assign(ESL.Media.create(), props);` or
`Object.assign(document.createElement('esl-media'), props);`

PR #2730, opened 2024-10-29
site: rework 3rd level nav, split complex doc

ala-n:


src/modules/esl-event-listener/docs/1-overview.md
```diff
@@ -0,0 +1,202 @@
+## Basic concepts
+
+The ESL event listener module is based on the following major terms and classes:
```
NastaLeo, line 3: a little bit strange to see ESL event listener written separately)

NastaLeo (approved):

src/modules/esl-event-listener/docs/3-extended-targets.md
```diff
@@ -0,0 +1,230 @@
+## <a name="extended-event-targets">Extended `EventTarget`s and standard optimizations</a> <i class="badge badge-sup badge-warning">beta</i>
```
abarmina, line 1: do we still need this anchor link here (and below in the 4th file)?

abarmina (approved):

PR #2748, opened 2024-11-06
chore(site): add share example

### LINKED ISSUE #2154: [🚀site]: migrate `esl-share` example page to playground base
As an ESL site holder, I want to have the [`esl-share`](https://esl-ui.com/examples/share/) page to be migrated to the UIPlayground base.

The current page is not mandatory proposed to be saved as a draft page.

yadamskaya:
ESL Share component has been migrated to UI Playground
@exadel-inc/esl-core-team please let me know if any additional component usage examples I need to add
Closes: https://github.com/exadel-inc/esl/issues/2154

### DIFF
--- a/site/views/examples/share.njk
+++ b/site/views/examples/share.njk
@@ -3,7 +3,7 @@ layout: content
 title: Share
 name: Share
 seoTitle: ESL Share - custom element to trigger various share actions for page or url with metadata based on ESL web components
-tags: [examples, beta]
+tags: [examples]
 icon: examples/share.svg
 aside:
   components:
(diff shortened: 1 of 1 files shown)


site/views/examples/share.njk
```diff
@@ -3,7 +3,7 @@ layout: content
 title: Share
 name: Share
 seoTitle: ESL Share - custom element to trigger various share actions for page or url with metadata based on ESL web components
-tags: [examples, beta]
+tags: [examples]
```
abarmina, line 6:
```suggestion
tags: [examples, playground]
```

site/views/examples/share.njk
```diff
@@ -20,59 +20,91 @@ aside:
...
+              uip-snippet-js="js-snippet-share">
+        <h4>Share button with default icon rendering</h4>
+        <esl-share-button name="linkedin" default-icon></esl-share-button>
+      </script>
 
-  <h3>Share button examples</h3>
-  <p>Example of share button with default icon rendering.</p>
-  <esl-share-button name="linkedin" default-icon></esl-share-button>
-  <hr>
+      <script type="text/html"
```
abarmina, line 34: maybe we can merge this with the previous tab? call it "Share button" and add the example without the icon that was there before? leave all 4 examples

image

site/views/examples/share.njk
ala-n: Little update:  we no longer use playground markers. I hope I'd even remove the renderer for playground

site/views/examples/share.njk
yadamskaya: @abarmina I like this idea, thanks

Issue #2751, opened 2024-11-06
[🚀esl-popup]: optimize popup position calculations and rethink offsetTrigger param


dshovchko:
There are two tasks for refactoring:
 - optimize calculations of popup positioning
 - rethink `offsetTrigger` param. Probably should be replaced with an attribute that specifies the distance between the popup and the trigger boundary

PR #2802, opened 2024-12-03
ESLPopup: add offset-trigger attribute

dshovchko:
Closes: #2751


NastaLeo (approved):

ala-n (approved):

ala-n:
TODO: discuss popup logic of exact overriding element attributes (probably we don't need it)
 - It is logical for sure to have the current options priority:
``` 
   opt = {
      ...STATIC_DEFAULTS,
      ...attributes,
      ...actionParams
   }
```
 - But it doesn't sound excellent that having actionProps defined leads to attributes being rewritten... This point may become a blocker for consumers...

fshovchko (approved):

Issue #2158, opened 2024-01-08
[🚀site]: migrate `esl-select`/`esl-select-list` example page to playground base


ala-n:
As an ESL site holder, I want to have the [`esl-select`](https://esl-ui.com/examples/forms-select/) and [`esl-select-list`](https://esl-ui.com/examples/forms-select-list/) page to be migrated to the UIPlayground base.

The current pages needs to be redirected to the new combined one https://esl-ui.com/examples/form/select/.
In case examples are too long and not representative propose to save them under drafts.

PR #2843, opened 2024-12-23
 Migrate `esl-select` and `esl-select-list` to uip

Natalie-Smirnova:
Closes: #2158


site/views/examples/forms-select-list.njk
```diff
@@ -3,11 +3,11 @@ layout: content
 title: Forms - Select List
 seoTitle: Custom select list examples based on ESL web components
 name: Select List
-tags: [examples, beta]
+tags: [examples, playground]
```
ala-n, line 6:
```suggestion
tags: [examples]
```

+ outdated with main-beta

site/views/examples/forms-select-list.njk
```diff
@@ -3,11 +3,11 @@ layout: content
...
 icon: examples/select-list.svg
 aside:
-  components:
-    - esl-select-list
+components:
+- esl-select-list
```
ala-n, line 10: Incorrect, due to front etter is yml syntax the offsets are important
Before (in JSON):
```js
   aside: {
      components: ['esl-select-list']
   }
```
Now (in JSON):
```js
aside: ''
components: ['esl-select-list']
```
Or to be precise incorrect even by syntax (mix list/map declaration)

site/views/examples/forms-select-list.njk
```diff
@@ -18,319 +18,74 @@ aside:
...
+      <script type="text/html"
+              label="Custom"
+              uip-snippet
+              uip-isolated
+              uip-snippet-js="js-snippet-select-list-registration">
+        <form class="form bg-white border p-4"  style="min-height: 150px">
+          <div class="form-group">
+            <label for="color_field" class="form-label">Choose your favorite colors:</label>
+            <esl-select-list class="border"  style="display: block">
               <select esl-select-target
```
ala-n, line 31: We should have at least one example with a long list for performance testing

ala-n (changes requested):
There is 2 additional big problems at that moment:
1) The examples are too short in size so dropdown can't be shown correctly
    We need to make playground area bigger or consider to use nonisolated scope (yes I was suggesting an isolated type, but due to presentation in convenience we probably should fallback to usual demo)
2) Usually the full demo (when all aspects of its implimenation) is cool, but seeing the result from user perspective I' afraid to see so much stub logic for displaying form data. Probably I was wrong here as well and we need to hide this presentation script. Plus one point to use non isolated scope (Yep sorry again, that was my suggestion, but I'm see that I was wrong) 


PS: due to holidays progress may be set on hold, it is not blocking 5.0.0 release and initial 5.0.0 site updates.

ala-n (changes requested):
@Natalie-Smirnova, please take a look at the validation check. 
PR is currently not compliant with a repository style guide 

PS: make sure you have husky(git hooks) installed and working locally, it seems it should prevent you from making commit

Issue #2879, opened 2025-01-23
[🐛 in esl-carousel]: CLS degradation for looped carousels with default renderer


ala-n:
**To Reproduce**
1. Create a page with a carousel component. Use loop mode, more than 2 slides, and default renderer.
2. Use heavy-weighted slides, e.g. background video (esl-media).
3. Check performance stats, see Cumulative Layout Shift.

**Current behavior**
There is a noticeable CLS (equal to the carousel percentage size) 

**Expected behavior**
ESL carousel should not produce CLS degradation.

ala-n:
Issue still reproducible in some cases on the prod usage, regression page is in progress...

PR #2880, opened 2025-01-23
fix(esl-carousel): fix initial rendering layout shift to eliminate CLS degradation

ala-n:
Closes: #2879

### DIFF
--- a/src/modules/esl-carousel/renderers/esl-carousel.default.renderer.less
+++ b/src/modules/esl-carousel/renderers/esl-carousel.default.renderer.less
@@ -8,9 +8,9 @@
   }
 
   &.esl-carousel-horizontal [esl-carousel-slide] {
-    width: var(--esl-slide-size);
+    width: var(--esl-slide-size, --esl-slide-fallback-size, 100%);
   }
   &.esl-carousel-vertical [esl-carousel-slide] {
-    height: var(--esl-slide-size);
+    height: var(--esl-slide-size, --esl-slide-fallback-size, 100%);
   }
 }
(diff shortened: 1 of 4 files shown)


NastaLeo (approved):

yadamskaya (approved):

abarmina (approved):

src/modules/esl-carousel/renderers/esl-carousel.default.renderer.less
```diff
@@ -8,9 +8,9 @@
   }
 
   &.esl-carousel-horizontal [esl-carousel-slide] {
-    width: var(--esl-slide-size);
+    width: var(--esl-slide-size, --esl-slide-fallback-size, 100%);
```
dshovchko, line 11: It seems to me that the user should use this variable `--esl-slide-fallback-size` to preset the slide intrinsic size. But the name sounds strange...

dshovchko (approved):

src/modules/esl-carousel/renderers/esl-carousel.default.renderer.less
ala-n: This size will be used until carousel initialised (no JS) so then it fallback (non primary source)

ala-n: As an idea `--esl-slide-initial-size`

dshovchko: I like it better this way.

Issue #2912, opened 2025-02-03
[🐛 in esl-media]: esl-media still indicates itself active when the youtube video is stopped


ala-n:
Steps to reproduce:

1. Open the test page (http://localhost:3005/examples/embedded-media/ or http://esl-ui.com/examples/embedded-media/)
2. Set any youtube video to src field
3. Observe the active marker of the video (dev console) or 
```html
  <button class="btn btn-sec-blue" esl-media-contol="{target: '::closest(uip-root)::find(.esl-media)', action: 'play', activeCls: 'btn-sec-green'}">Play
  </button>
```
When https://github.com/exadel-inc/esl/pull/2878 will be released or available
4. Start video then Stop (not pause)

Current behavior: 
 - component still indicates itself as an active

Expected behavior:
 - component should not be marked as active

PR #2918, opened 2025-02-05
fix(esl-media): detect youtube provider stopped state

fshovchko:
Closes: #2912

The issue was caused by an interesting behavior of YouTube api:
Basically we were expecting for `PlayerState.ENDED` event to be triggered once video is stopped. But it doesn't. It only happens when video was watched until the end and then automatically stopped.

Behavior is different with a manual stop:
The behavior i got was two event triggered a stop - `PlayerState.UNSTARTED` (-1) and `PlayerState.CUED` (5).

And the [YT API](https://developers.google.com/youtube/iframe_api_reference#Playback_controls) says, that the stopVideo function could put the player into any not-playing state, including ended (0), paused (2), video cued (5) or unstarted (-1).

My guess that unstarted (-1) is triggered every time when video is at 0 seconds and shows sort of the 'preview' state.

I couldn't find any good proposals to address the issue, so decided if player was in playing or paused state and then abruptly put in unstarted state, it means it was manually paused

### DIFF
--- a/src/modules/esl-media/providers/youtube-provider.ts
+++ b/src/modules/esl-media/providers/youtube-provider.ts
@@ -20,7 +20,7 @@ export class YouTubeProvider extends BaseProvider {
   protected override _el: HTMLDivElement | HTMLIFrameElement;
   protected _api: YT.Player;
 
-  protected lastPlayerState: null | number;
+  protected _lastPlayerState: PlayerStates = PlayerStates.UNINITIALIZED;
 
   static override parseUrl(url: string): Partial<MediaProviderConfig> | null {
     if (this.providerRegexp.test(url)) {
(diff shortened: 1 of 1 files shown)


src/modules/esl-media/providers/youtube-provider.ts
```diff
@@ -20,6 +20,8 @@ export class YouTubeProvider extends BaseProvider {
   protected override _el: HTMLDivElement | HTMLIFrameElement;
   protected _api: YT.Player;
 
+  protected lastPlayerState: null | number;
```
ala-n, line 23: Cosmetic: 
```suggestion
  protected _lastPlayerState: PlayerState = PlayerState.UNINITIALIZED;
```

ala-n (approved):
Actually, I like your solution. That definitely works.

dshovchko (approved):

yadamskaya (approved):

abarmina (approved):

Issue #2959, opened 2025-02-27
[🐛 in esl-scrollbar]: incorrect drag behavior when targeting <html/> tag


fshovchko:
**To Reproduce**
1. Go to https://esl-ui.com/dev/nested-accordion/
2. Drag a scrollbar

**Current behavior**
Scroll dragging instantly snaps either to start or to end of the page

**Expected behavior**
Smooth drag behavior

PR #2986, opened 2025-03-12
refactor(site): introduce single scrollbar for all page layouts

fshovchko:
In the scope of this PR, the scrollable element of all esl site layouts was changed from `main` to `html`, which simplified templates and styles and also made browsers remember scroll position which didn't work previously.

Relates to: #2959


ala-n:
🐛 (Normal): On XS, opened sidebar does not lock page scroll -> can cause unexpected behaviour on mobile. Probably we just need add ScrollLock util to sidebar.

ala-n:
🐛 (Normal): Collapsing artifact during hamburger menu/sidebar opening animation. It seems to be content allowed to be shrink which is not good page behaviour. Suggesting making mobile menu out of page flow (position fixed)

ala-n (commented):
No code concerns, all looks cool 👍🏻 
Just two issue to resolve for XS/Mobile

dshovchko (approved):

fshovchko:
@ala-n, additional 🐛 was found with a scrollbar disappearing on small size landing page. It was due benefits list having css setting margin; 0 - 20px. It was causing scrollbar to shift outside of window during animations

ala-n (approved):
✅ Visual Testing Passed as well

dshovchko (approved):

NastaLeo (approved):

Issue #2997, opened 2025-03-17
[🚀e2e tests]: update snapshot-tests package to use ESM/TS independent output


ala-n:
Requested to make `snapshot-tests` subpackage as pure ESM module.
Handle dependencies:
 - e.g. https://github.com/exadel-inc/esl/pull/2949 accordingly

PR #3030, opened 2025-04-01
chore(e2e): update package to ues ts and esm output

fshovchko:
Closes: #2997


packages/snapshot-tests/jest-puppeteer.config.js
```diff
@@ -2,8 +2,10 @@ if (!process.argv.includes('--no-autorun') && !process.env.PORT) {
   process.env.PORT = '3007';
 }
 
-/** @type {import('jest-environment-puppeteer').JestPuppeteerConfig} */
-module.exports = {
+const {JestPuppeteerConfig} = await import('jest-environment-puppeteer');
+
+/** @type {JestPuppeteerConfig} */
```
ala-n, line 7: We should not make actual and moreover async imports for JS Doc type hint

packages/snapshot-tests/jest.config.js
```diff
@@ -1,17 +1,23 @@
...
-const rimraf = require('rimraf');
 rimraf.sync('./.diff');
 
 // Actual Jest configuration
-module.exports = {
+export default {
   preset: 'jest-puppeteer',
   transform: {
-    '^.+\\.tsx?$': 'ts-jest',
+    '^.+\\.(j|t)sx?$': '@swc/jest',
```
ala-n, line 14: We do not support JS or JSX sources. Our code base should be TS as much as possible.

packages/snapshot-tests/jest.config.js
```diff
@@ -1,17 +1,23 @@
...
     '^.+\\.feature$': './src/transformer/gherkin.js'
   },
+  transformIgnorePatterns: [
+    "/node_modules/(?!pixelmatch)"
+  ],
+  extensionsToTreatAsEsm: ['.tsx', '.jsx'],
```
ala-n, line 20:
```suggestion
  extensionsToTreatAsEsm: ['.ts', '.tsx'],
```

packages/snapshot-tests/jest.config.js
```diff
@@ -1,17 +1,23 @@
...
+export default {
   preset: 'jest-puppeteer',
   transform: {
-    '^.+\\.tsx?$': 'ts-jest',
+    '^.+\\.(j|t)sx?$': '@swc/jest',
     '^.+\\.feature$': './src/transformer/gherkin.js'
   },
+  transformIgnorePatterns: [
+    "/node_modules/(?!pixelmatch)"
+  ],
```
ala-n, line 19: ?? Why do we need this exclude it seems like we started from pixelmatch updated to ESM ...

packages/snapshot-tests/jest.config.js
```diff
@@ -1,17 +1,21 @@
...
+// Enable ESM modules in Jest
+process.env.NODE_OPTIONS = '--experimental-vm-modules';
 
 // Actual Jest configuration
-module.exports = {
+export default {
   preset: 'jest-puppeteer',
   transform: {
-    '^.+\\.tsx?$': 'ts-jest',
+    '^.+\\.tsx?$': '@swc/jest',
```
ala-n, line 15: Probably, after all the other modifications we have, the global `ts-jest` will work here:

```suggestion
    ['ts-jest', { useESM: true }],
```

I didn’t properly clarify my point earlier (on weakly meeting), sorry about that. What I meant was: if swc solves our problems or at least simplifies something with ESM, then I’m open to using it here. Otherwise, there’s no real benefit to using it.

The only general advantage of swc is its speed. However, in this case, we’re not focusing on performance, and it won’t provide any noticeable improvements. 
PS: We can discuss it on a call detailed just in case.

ala-n (commented):
Other then question with transformer and resolving conflicts looks good 👍🏻

packages/snapshot-tests/jest.config.js
fshovchko: I don't insist on using swc. The only reason it was utilized is because tests were failing in any other configuration. But now it does, even though I was removing cache every single time.

NastaLeo (approved):

dshovchko (approved):

PR #3105, opened 2025-05-16
epic: massive project build process rework + CI fixes

ala-n:
**Key points in this PR:**

1. **Build process migrated to Nx targets**
   The entire build process has been moved to Nx targets. Nx provides caching, build planning, and parallelization. Configuration for each submodule is now stored in `project.json` files under Nx.

2. **Simplified npm scripts and unified command usage**
   The `npm` scripts now represent only the public API. There are no sub-steps defined in `package.json`.
   Regarding Lerna vs Nx: since Nx is the actual build engine, we use Nx commands exclusively. With full Nx configuration, the commands are clear, short, and simple. There is no restriction to run lerna commands but it is nofing else but original nx commands under the hood.

3. **Partial use of Nx-inferred tasks with custom adjustments**
   We're partially using Nx-inferred tasks, but they don’t cover everything we need. A few specific points:

   * We do not use `nx:jest` to run e2e tests. Unfortinatly configuration capabilities of `nx:jest` sucks, we may even return for direct usage of jest for unit tests in future.
   * We import `eslint-plugin-esl` asynchronously and safely because `nx:eslint` analyzes ESLint config files without executing prebuild tasks (to build `eslint-plugin-esl`).

4. **Removed `concurrently` dependency**
   We no longer need `concurrently`, as Nx supports everything it provided.
   **Note:** Don't rely on AI tools (ChatGPT, Copilot, etc.)—they often misunderstand Nx's capabilities. That said, Nx documentation for developers is also very far from ideal.

5. **Simplified substeps**
   Some subtasks have been merged for simplicity (even at the cost of slightly longer build times). For example, the CSS minification is now part of the LESS build process.

6. **Renamed snapshot testing module**
   The `snapshot-tests` module has been renamed to `esl-e2e`. This is the only viable way of classification for the module within Nx.

7. **Shortened package names for Nx**
   Package names have been shortened—there's no need to include the scope (`@exadel`) part of the package name when using Nx commands.

8. **Extended and improved watch process**
   The watch process has been reworked. One of the most important tricks is related to using `.done` files generated by ESL library build processes. These empty `.done` files update their modification date at the end of the build, which allows effective change detection without prematurely triggering dependent processes.


AlexanderBazukevich (approved):

abarmina (approved):

NastaLeo (approved):

yadamskaya (approved):

dshovchko (approved):

package.json
```diff
@@ -48,12 +50,14 @@
     "kleur": "^4.1.5",
     "lerna": "^8.2.2",
     "less": "^4.3.0",
+    "less-plugin-clean-css": "^1.6.0",
     "lint-staged": "^16.0.0",
     "postcss": "^8.5.3",
```
dshovchko, line 55: Do we need postcss?

PR #3403, opened 2025-09-26
test(e2e): move snapshot url handler to js

NastaLeo:


packages/snapshot-tests/src/reporters/reporter.js
```diff
@@ -18,6 +18,14 @@ export class SnapshotAwareReporter {
     this._totalStartTime = performance.now();
   }
 
+  get baseUrl() {
+    if (!process.env.GITHUB_ACTIONS || !process.env.DIFF_REPORT_BRANCH || !this._options.outputPublishPath) return null;
```
ala-n, line 22:
```suggestion
    if (!process.env.GITHUB_ACTIONS || !process.env.DIFF_REPORT_BRANCH) return null;
```

ala-n (approved):

Issue #3419, opened 2025-10-01
[🐛 ✏️ui-playground]: uip-settings & uip-select-setting theme support is broken


ala-n:
## Steps to reproduce
1. Go to [/playground/](https://esl-ui.com/playground/)
2. Open logo "Color" selector, observe items colors:
   - **Actual result**: the select dropdown in dark mode
   - **Expect result**:  the select dropdown in light mode when the environment is in light mode

3. Change the theme to dark mode
4. Observe settings sidebar
  - **Actual result**: the settings sidebar is still in the light mode
  - **Expect result**:  the setting sidebar is in the dark mode colors

PR #3443, opened 2025-10-14
feat(ui-playground): theme class attribute to ease theme override

fshovchko:
In the scope of this PR the way UIP theme is overridden on ESL was refactored using class.
But in order to do override correctly the class should be present not only on `uip-root`, but also on settings select dropdown, since it's an `esl-select` instance and appended outside of `uip-root`. That's why an attribute was introduced as well.

Closes: #3419

### DIFF
--- a/packages/esl-website/views/examples/anchornav.njk
+++ b/packages/esl-website/views/examples/anchornav.njk
@@ -97,7 +97,7 @@ aside:
 
       <uip-snippets class="uip-toolbar" dropdown-view="@xs"></uip-snippets>
       <uip-settings vertical="@+sm">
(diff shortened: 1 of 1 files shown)


ala-n (commented):
Regarding settings list not sure it is the best stratgy for us to relay on optional class (even if there is no concerns for it in general) I'd vote for style trnasfer to esl-select.

Additionally there is a couple of topics to consideration:
1) By some reason our latest version of esl-select is laggy on the start, we need to investiggate wy do we see at least one frame of dropdown popup that is not ready to be shown

2) As the web deweloped it might esl-select become legacy due too browser caapabilites. We need to investiggate it, so proposing to sart from UIP Select option and see if there is an easy way to migrate and use modern browser capabilities without loosing posibility to setup custom styling

dshovchko (approved):

PR #3545, opened 2025-12-17
test(deps-dev): replace `jest` with `vitest`

fshovchko:


packages/esl/src/esl-utils/test/timer.mock.ts
```diff
@@ -0,0 +1,78 @@
...
+    const timeout = realSetTimeout(() => {
+      // console.log('Timeout fired', id);
+      if (!timeouts.has(id)) return;
+      // console.log('setTimeout mock', id, timeout, callback.toString());
+      timeouts.delete(id);
+
+      if (typeof callback === 'function') {
+        callback(...args);
+      } else {
+        eval(callback);
```
dshovchko: Not sure we need eval of string representation of handler

PR #3833, opened 2026-05-28
feat(esl-match-height): introduce ESLMatchHeightMixin

fshovchko:
In the scope of the PR the `ESLMatchHeight` (`esl-match-height`) is introduced. It's a custom mixin element that equalizes heights of child elements within a container.
It automatically groups elements by their vertical position (row) and sets each group to the maximum height found in that group.


packages/esl/src/esl-match-height/README.md
```diff
@@ -0,0 +1,68 @@
...
+  <div class="card">Regular card</div>
+</div>
+```
+
+### Behavior
+
+- On initialization, the mixin queries all matching child elements and equalizes their heights per row.
+- The mixin automatically recalculates heights on container resize and font load events.
+- Height updates are throttled (200ms) for performance.
+- When disconnected, the mixin clears all applied heights.
```
dshovchko, line 61: Looks like this point is not realized, and tests do not cover the declared feature.

dshovchko (approved):

packages/esl/src/esl-match-height/README.md
fshovchko: Good catch, I removed the disconnected callback feature but didn't update the docs

ala-n (approved):

packages/esl/src/esl-match-height/core/esl-match-height.ts
```diff
@@ -0,0 +1,118 @@
...
+  /** List of HTMLElements to normalize height */
+  public get $elements(): HTMLElement[] {
+    return Array.from(this.$host.querySelectorAll(this.selector || ESLMatchHeightMxin.DEFAULT_SELECTOR));
+  }
+
+  protected override connectedCallback(): void {
+    super.connectedCallback();
+    afterNextRender(() => this.update());
+  }
+
```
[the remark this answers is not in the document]
ala-n: For that component library level, we definitely need the way to revert things as soon as the plugin is disabled (removed from the host)
It also sounds reasonable to handle attribute change

NTH for future support for the ESLMediaRuleList list for the main attribute value.

packages/esl/src/esl-match-height/core/esl-match-height.ts
```diff
@@ -0,0 +1,118 @@
...
+  /** Compares two {@link MatchItem} objects */
+  @bind
+  public compare(a: MatchItem, b: MatchItem): number {
+    if (a.order === b.order) return a.top - b.top;
+    return (a.order - b.order) * Number.POSITIVE_INFINITY;
+  }
+
+  /** Creates a wrapper object based on the passed element */
+  @bind
+  protected wrap($el: HTMLElement): MatchItem {
```
ala-n:
```suggestion
  protected toMatchItem($el: HTMLElement): MatchItem {
```

abarmina (approved):

NastaLeo (approved):

packages/esl/src/esl-match-height/core/esl-match-height.ts
```diff
@@ -0,0 +1,131 @@
...
+ *      esl-match-height-order="selectorTopPriority | selectorSecondPriority">
+ *   ... div.selector * n
+ * </div>
+ * ```
+ */
+@ExportNs('MatchHeight')
+export class ESLMatchHeightMixin extends ESLMixinElement {
+  public static override is = 'esl-match-height';
+
+  static override observedAttributes = [ESLMatchHeightMixin.is, 'esl-match-height-order'];
```
ala-n:
```suggestion
  static override observedAttributes = ['esl-match-height-order'];
```
Mixin always obseves its primary attribute

ala-n (approved):

ala-n (changes requested):
Unfortunately, the component did not pass live testing
@fshovchko, please review the updated example page with a realistic case for that plugin

ala-n (approved):

dshovchko (approved):

NastaLeo (approved):

PR #3887, opened 2026-06-23
chore: fix npx skills command sample, add skill version, update build scripts, fix flaky tests 

ala-n:


ala-n:
@exadel-inc/esl-core-team, please review this small update post facto
Just don't want to overbother you with the quick review requests of small build updates and bugfixes

PR #3936, opened 2026-07-16
fix(esl-utils): ESLVSizeCSSProxy API updated + html element size observation added to handle scroll changes + proper singleton guard

ala-n:


packages/esl/src/esl-utils/fixes/viewport.ts
```diff
@@ -1,31 +1,93 @@
...
-  /** Destroys observer */
+  /** Stops viewport observation for the singleton proxy instance, removes CSS vars and cancels a pending frame. */
   public static destroy(): void {
-    window.removeEventListener('resize', this.updateDebounced);
+    new ESLVSizeCSSProxy().unobserve();
+  }
+  /** Schedules CSS variables refresh for the singleton proxy instance. */
+  public static update(): void {
+    new ESLVSizeCSSProxy().requestUpdate();
+  }
```
dshovchko: Looks like it would work after `destroy()` or without `init()`.
Definitely, this is manual update trigger, but with two nuances:
 - this is not forced write CSS variables but write if changed (because of guard)
 - this is unnecessary almost when active init() (only when size changed in a way that neither `window.resize` nor `ResizeObserver` on <html> catch)

Guess we should document this.

dshovchko (approved):

fshovchko (approved):

dshovchko (approved):

abarmina (approved):

PR #4003, opened 2026-08-19
epic: pure CSS output (wave 1)

### LINKED ISSUE #4004: [🔧 build]: migrate ESL component styles sources and output from LESS to pure CSS
Migrate `@exadel/esl` module to pure CSS output and sources.

Main tasks
 - Convert each remaining module’s LESS sources and public entrypoints to CSS.
 - Preserve the existing public style entrypoint structure (core.css / all.css).
 - Replace LESS-only syntax with CSS-compliant equivalents.
 - Ensure component CSS imports are resolved by PostCSS.
 - Migrate or remove LESS mixin dependencies in esl-image, esl-media, esl-popup, and esl-scrollbar.

Modules:
 - [x] esl-a11y-group
 - [x] esl-alert
 - [x] esl-animate
 - [x] esl-anchornav
 - [x] esl-drag-to-scroll
 - [x] esl-line-clamp
 - [x] esl-sticky-box

 - [x] esl-utils

 - [x] esl-image-utils
 - [x] esl-scrollbar

 - [x] esl-image
 - [x] esl-media

 - [x] esl-toggleable
 - [x] esl-popup
 - [x] esl-trigger
 - [x] esl-panel
 - [x] esl-panel-group
 - [x] esl-tab
 - [x] esl-forms
 - [x] esl-footnotes
 - [x] esl-tooltip
 - [x] esl-share

 - [x] esl-carousel

 - [x] Replace the temporary leaf-style imports in src/all.less once the all-in-one bundle is migrated from the Less pipeline.

ala-n:
Relates to: https://github.com/exadel-inc/esl/issues/4004


NastaLeo (approved):

packages/stylelint-config/custom/import.js
```diff
@@ -2,10 +2,7 @@ import stylelint from 'stylelint';
 
 const {validateOptions, ruleMessages, report} = stylelint.utils;
 
-const trimPath = (path) => path
-  .replace(/^\s*(url\()?['"]?/, '')
-  .replace(/['"]?\)?\s*$/, '')
-  .trim();
+const trimPath = (path) => path.replace(/^\s*(?:\(\w+\))?['"\s]*|['"\s]*$/g, '');
```
dshovchko, line 5: Is it the correct replacement? It seems to me that url() syntax doesn't work for processing.

```suggestion
const trimPath = (path) => path.replace(/^\s*(?:url\()?['"]?|['"]?\)?\s*$/g, '');

```

dshovchko (approved):

packages/stylelint-config/custom/import.js
ala-n: The problem is that it disallows the `(inline)` import type as well
Practically, that's not even matter, as this rule will be restricted as soon as we move to CSS

PR #4017, opened 2026-08-27
feat(esl-media): migrate module styles to pure CSS export

ala-n:
- remove `core.mixin.less` LESS mixin export

BREAKING CHANGE: `esl-media` styles are now exported as pure CSS. Use `@exadel/esl/modules/esl-media/core.css` instead of LESS mixins/imports (`core.mixin.less`, `core.less`).


dshovchko (approved):

NastaLeo (approved):

packages/esl/src/esl-media/core/esl-media-container.css
```diff
@@ -0,0 +1,7 @@
+.img-container > esl-media {
+  position: absolute;
+  top: 0;
```
abarmina, line 3: nit: can be just `inset: 0`, isn’t it?

abarmina (approved):

packages/esl/src/esl-media/core/esl-media-container.css
ala-n: It is not 100% the same unfortunately, it will be dangerous change without extra testing

PR #4053, opened 2026-09-11
[🔧 ts compatibility]: support accessors as decorator members

### LINKED ISSUE #3927: [🔧 ts compatibility]: support `@attr`, `@jsonAttr`, `@boolAttr`, `@prop` for `accessor` class members
The new `accessor` declaration was introduced in TypeScript 5+.

As an ESL consumer, I want to be able to use ESL decorators with `accessor` class members.

### Example

```ts
class MyClass {
  @attr()
  public accessor myAttr;
}
```

Currently, ESL decorators do not support `accessor` members.

### Expected behavior

ESL decorators should recognize and correctly decorate `accessor` class members in the same way as other supported class members.

The implementation should preserve the expected accessor behavior and remain compatible with the existing decorator API.

fshovchko:
Closes: #3927

### DIFF
--- a/packages/esl/src/esl-utils/decorators/test/accessor.test-d.ts
+++ b/packages/esl/src/esl-utils/decorators/test/accessor.test-d.ts
@@ -1,0 +1,32 @@
+import {attr} from '../attr';
+import {boolAttr} from '../bool-attr';
+import {jsonAttr} from '../json-attr';
+import {prop} from '../prop';
+
+class AccessorTypesTest extends HTMLElement {
+  @attr()
+  public accessor attrAccessor: string;
+  @attr({defaultValue: 'def'})
+  public accessor attrAccessorDef: string;
+  @attr({readonly: true})
+  public accessor attrAccessorReadonly: string | null;
+
+  @boolAttr()
+  public accessor boolAccessor: boolean;
+
+  @jsonAttr()
+  public accessor jsonAccessor: Record<string, unknown>;
+
+  @prop('value')
+  public accessor propAccessor: string;
+
+  // Plain fields keep working as before
+  @attr()
+  public attrField: string;
+  @boolAttr()
+  public boolField: boolean;
+  @jsonAttr()
+  public jsonField: Record<string, unknown>;
+  @prop('value')
+  public propField: string;
+}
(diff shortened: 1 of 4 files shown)


packages/esl/src/esl-utils/dom/attr.ts
```diff
@@ -3,7 +3,7 @@ import {resolveDomTarget} from '../abstract/dom-target';
 import type {ESLDomElementTarget} from '../abstract/dom-target';
 
 export type ESLAttributeTarget = undefined | null | ESLDomElementTarget | ESLDomElementTarget[];
-export type ESLAttributeDecorator = (target: ESLDomElementTarget, propName: string) => void;
+export type ESLAttributeDecorator = (target: ESLDomElementTarget, propName: string, descriptor?: PropertyDescriptor) => void;
```
dshovchko, line 6: The type now optionally declares descriptor parameter but none of the concrete decorator factories (attr, boolAttr, jsonAttr) accept or use it. This is not functional defect but a matter of contract clarity.

```suggestion
/**
 * Attribute decorator compatible with fields and auto-accessors.
 * The incoming accessor descriptor is intentionally ignored because
 * attribute decorators replace it with their own property descriptor.
 */
export type ESLAttributeDecorator = (target: ESLDomElementTarget, propName: string, descriptor?: PropertyDescriptor) => void;
```

dshovchko (approved):

packages/esl/src/esl-utils/decorators/test/accessor.test.ts
```diff
@@ -0,0 +1,101 @@
+import {attr} from '../attr';
+import {boolAttr} from '../bool-attr';
+import {jsonAttr} from '../json-attr';
+import {prop} from '../prop';
+
+describe('Decorators: accessor class members support', () => {
```
ala-n, line 6: It's better to split this to `${submodule}.accessor.test.ts` (e.g. `prop.accessor.test.ts`) files or append to existing

packages/esl/src/esl-utils/decorators/test/accessor.test-d.ts
```diff
@@ -0,0 +1,32 @@
+import {attr} from '../attr';
+import {boolAttr} from '../bool-attr';
+import {jsonAttr} from '../json-attr';
+import {prop} from '../prop';
+
+class AccessorTypesTest extends HTMLElement {
```
ala-n, line 6: Same with type tests if we really need them, prefer `submodule.feature.test-d.ts` notation for file and code splitting

